PR #14, opened 2015-03-23
Add docs

Ladsgroup:
Use Sphnix for auto-documenting. Need some work specially in dependencies.
Also add some documentation as an example


halfak:
I tried "make html" and got

Theme error:
no theme named 'alabaster' found (missing theme.conf?)

Ladsgroup:
This is strange since the theme is a built-in theme.
http://sphinx-doc.org/theming.html
Anyway I changed it to default and tested it in another clone, worked okay.

halfak:
Solid.  {{merged}}

PR #36, opened 2015-05-26
Adds configuration for uwsgi+nginx 

halfak:
Also, implements batch style scoring which should provide a substantial speed improvement.

### DIFF
--- a/config/ores.yaml
+++ b/config/ores.yaml
@@ -1,0 +1,85 @@
+application_root: "/scores"
+url_prefix: "/scores"
+
+# Scorers
+scorers:
+  enwiki:
+    extractor: enwiki_api
+    models:
+      reverted: enwiki_revert
+  ptwiki:
+    extractor: ptwiki_api
+    models:
+      reverted: ptwiki_revert
+  fawiki:
+    extractor: fawiki_api
+    models:
+      reverted: fawiki_revert
+#  trwiki:
+#    extractor: trwiki_api
+#    models:
+#      reverted: trwiki_revert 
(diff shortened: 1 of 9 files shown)


config/ores.yaml
```diff
@@ -0,0 +1,85 @@
...
+  fawiki_api:
+    url: https://fa.wikipedia.org/w/api.php
+    language: persian
+  trwiki_api:
+    url: https://tr.wikipedia.org/w/api.php
+    language: turkish
+  frwiki_api:
+    url: https://fr.wikipedia.org/w/api.php
+    language: french
+  simplewiki_api:
```
he7d3r: Should this be commented out too, as above?

config/ores.yaml
```diff
@@ -0,0 +1,85 @@
...
+    model_file: models/ptwiki.reverted.linear_svc.model
+  frwiki_revert:
+    model_file: models/frwiki.reverted.linear_svc.model
+  fawiki_revert:
+    model_file: models/fawiki.reverted.linear_svc.model
+  ptwiki_revert:
+    model_file: models/ptwiki.reverted.linear_svc.model
+  trwiki_revert:
+    model_file: models/trwiki.reverted.linear_svc.model
+  simplewiki_revert:
```
he7d3r: same here

config/ores.yaml
halfak: Negative.  That way, once those models are ready, we only have to remove one set of comments.

he7d3r:
Looks good.

PR #38, opened 2015-05-27
Add way to configure custom NLTK data path

yuvipanda:
This allows us to manage where to download the data, which
is important when using a virtualenv for automated deploys

### DIFF
--- a/ores/wsgi/application.py
+++ b/ores/wsgi/application.py
@@ -1,13 +1,17 @@
 import json
 import os
 
-from flask import Blueprint, Flask, request
-from revscoring.scorers import Scorer
-
-from . import routes
-
 
 def configure(config):
+    if 'data_path' in config and 'nltk' in config['data_path']:
+        import nltk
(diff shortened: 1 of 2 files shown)


he7d3r:
Looks good to me.

ores/wsgi/application.py
```diff
@@ -1,13 +1,17 @@
...
-from . import routes
+def configure(config):
+    if 'data_path' in config and 'nltk' in config['data_path']:
+        import nltk
+        nltk.data.path.append(config['data_path']['nltk'])
 
+    from flask import Blueprint, Flask, request
+    from revscoring.scorers import Scorer
+
+    from . import routes
```
halfak: Why put these imports within the configure function?

yuvipanda: Because they otherwise import nltk and try to use it without having set the path and it fails. And I need the config object to set the path.

halfak: Oh!  I see.  Feels hacky to me, but I can't see how else it will work.  Merging.

PR #48, opened 2015-06-07
Adds makefile for building relevant models.

halfak:
Adds a Makefile for building intermediate datasets and model files.


he7d3r:
Looks good

PR #56, opened 2015-06-13
Moves makefile to ores-wikimedia-config

halfak:


halfak:
See also https://github.com/wiki-ai/ores-wikimedia-config/pull/6

PR #64, opened 2015-07-16
Removes feature_lists -- moved to ores-wikimedia-config

halfak:


halfak:
See https://github.com/wiki-ai/ores-wikimedia-config/pull/12

PR #69, opened 2015-07-21
Stub install documentation

adamwight:


halfak:
This is great.  It's definitely something we should flesh out.  The most burdensome part of the install requirements are related to the 'revscoring' package.  See https://github.com/wiki-ai/revscoring/blob/master/README.rst#installation  

It seems that there's a lot missing from your notes that are requirements for that package -- including the full set of aspell and myspell dictionary packages (see https://github.com/wiki-ai/revscoring/issues/138).  

There's also the terribleness of scipy and numpy's dependencies.  See https://github.com/wiki-ai/revscoring/issues/139

Given that this improves the non-existent installation documentation, I'm partial to just merging it if you'd like that I do, but I figured that I'd give you the chance to iterate on it and get it fleshed out before I do.

adamwight:
Aha, okay I'll tease out which are scipy/numpy dependencies, and which are direct revscoring dependencies.

What do you think about packaging as a .deb?  I just did a bit of that, and would be happy to repeat here.  It's mostly an issue of how you see this being used and deployed.  .debs are incredibly convenient for deployment, if you want to go that way, but perhaps slightly less convenient to release.

halfak:
Not sure how I feel about it since I've never wrapped something up in a .deb before.  I assume that, if we're wrapping up ORES in a deb, I need to wrap all of it's dependencies that aren't debs too?  E.g. 'revscoring', 'yamlconf', 'flask-jsonpify', etc.

adamwight:
Here's an example of the packaging required for a python module.  Let's pretend I didn't suggest this ;)

https://sources.debian.net/src/python-pip/1.5.6-6/debian/

yuvipanda:
#70 might make bits of these obsolete.

adamwight:
I'm happy to document whatever you think we should recommend!  Should we say that vagrant is the only sane way to set up a development environment, or virtualenv, or bare metal?  Or is it still possible to use any of these three methods?

I'm reserving my opinion on what is most sane, until we at least have documentation that covers however this is used in the real world...

yuvipanda:
I personally think vagrant should be the only supported dev environment, because it unifies underlying OS / versions for everyone - 'it worked on my machine' things become a lot less common. It also works on OS X / Windows / Linux the same, which is a huge advantage IMO. 

Do try out the vagrant patch and let me know if it works for you :)

yuvipanda:
It'll also closely match our production environment, reducing surprises caused by differing versions of things.

adamwight:
Hrm, DO NOT MERGE--please note that I accidentally included @yuvipanda 's #70 in this branch, it should be a dependency instead.

README.md
```diff
@@ -1,2 +1,32 @@
...
+## Or, the hard way!
+If you want to install by hand, follow the installation instructions for
+[revscoring](https://github.com/wiki-ai/revscoring/blob/master/README.rst#installation),
+especially the package installation.  Please note that your setup will probably not
+be perfectly compatible with our development and production platforms.  Consider
+using the Vagrant method, above.
+
+Install ORES like any other python module, for example:
+
+* `sudo pip3 install ores`
```
he7d3r: Is `sudo` required?

README.md
```diff
@@ -1,2 +1,32 @@
 # Objective Revision Evaluation Service
 A webserver for hosting scoring services.
+
+# Installation
+
+## Vagrant
+The recommended installation is to set up Vagrant and VirtualBox, then `cd ores` and run:
+
+* `vagrant up`
```
halfak: Should have step 1 and 2 as "install virtualbox" and "install vagrant" with links to their docs.

PR #75, opened 2015-07-29
Fixes issues with vagrant startup.  

halfak:
1. Virtualbox won't try to use USB -- no good reason we need that.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,4 +9,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision "shell", path: 'provision.bash'
 
   config.vm.network "forwarded_port", guest: 8080, host: 8080
+
+  config.vm.provider "virtualbox" do |vb|
+    vb.customize ["modifyvm", :id, "--usb", "off"]
+    vb.customize ["modifyvm", :id, "--usbehci", "off"]
+  end
 end
+
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ celery
 docopt
 flask
 flask-jsonpify
-mediawiki-utilities==0.4.15
+mediawiki-utilities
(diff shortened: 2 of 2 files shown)


halfak:
@adamwight, could you take a look at this one?

adamwight:
So far, I'm getting success on the first `vagrant up`, using the master branch.

There are some errors compiling numpy, "libyaml is not found" it looks bad but it's not causing the overall provisioning to fail, so I guess that's recoverable.

Can you explain what problem this is fixing?  Have any error messages to paste here?

adamwight:
Looks helpful, merging!

I was unable to reproduce the bug fixed here.

halfak:
Sorry for the delay, here's the error message I got from Vbox:

```
Failed to open a session for the virtual machine .... Implementation of the USB 2.0 controller not found! Because the USB 2.0 controller state is part of the saved VM state, the VM cannot be started. To fix this problem, either install the 'Oracle VM VirtualBox Extension Pack' or disable USB 2.0 support in the VM settings (VERR_NOT_FOUND). Unknown error creating VM (VERR_NOT_FOUND).
```

This is remedied by installing the 'Oracle VM VirtualBox Extension Pack', but I thought it was a better idea to just disable USB 2.0 support.

adamwight:
Ah, that makes sense, thanks for finding a workaround!  Yeah, I had the extension pack enabled already therefore didn't experience the error, but it's not installed by default.

halfak:
@adamwight  merge-y?

adamwight:
Hah,
> Pull requests that have a failing status can’t be merged from a phone.

Whoops, gimme a few minutes. Thought I had already merged this!

adamwight:
Hah,
> Pull requests that have a failing status can’t be merged from a phone.

Whoops, gimme a few minutes. Thought I had already merged this!

PR #78, opened 2015-08-08
Celery performance improvements

halfak:
* Configures celery state to update to 'SENT' correctly.
* Implements special case processor for a single revision score that Improves simultaneous request performance.


yuvipanda:
Have you tested this anywhere? Staging? Locally?

halfak:
Works locally and I confirmed the performance gains for running many precached.  There's still only a 90% hit rate when all precached are run with no delay, but that's way better than 0% we get now.

PR #82, opened 2015-09-05
Updates for revscoring refactoring.

halfak:
Removes explicit usage of 'language' as a context.


he7d3r:
Tests are failing for me with this patch applied:
> ERROR: Failure: ImportError (cannot import name 'RevisionNotFound')
http://dpaste.com/2KQTY70

halfak:
You need to install the new version of revscoring.

he7d3r:
Indeed. They are passing now.

PR #83, opened 2015-09-09
Simplify provisioner script

yuvipanda:


halfak:
Tested on my local machine

vagrant destroy
vagrant up

worked as expected

Issue #94, opened 2015-09-30
Basic request metrics collection and a magic word for precached


halfak:
We should have basic logging that counts (1) the number of scores requested per amount of time, (2) the proportion of requests that are returned from the cache, generated or errored, and (3) the response time of 

Statsd seems like a good option.  

We'll need a magic word to flag requests coming from precached to be logged differently so that we don't count those as real scores used.  

This will also need to be abstracted in some way so that such logging can be optional.

halfak:
Fixed by https://github.com/wiki-ai/ores/pull/96

PR #96, opened 2015-10-01
Adds support for metrics collection

halfak:


ToAruShiroiNeko:
I think it would be useful to take model version number in to account when generating statistics. We can use those as cut off points or defacto milestones.

ToAruShiroiNeko:
-  score_processor: local_celery
+  score_processor: local_timeout

What ıs the reason for this change? I do not uite understand it.

Aside from that it looks good to me.

halfak:
@ToAruShiroiNeko good idea.  Versions are now recorded.  

Regarding the change from local_celery to local_timeout, that makes setting up the dev server a little easier because you don't need a celery worker instance to run it.  I used that in testing to make sure we were logging metrics correctly.

halfak:
Fixes https://github.com/wiki-ai/ores/issues/94

PR #102, opened 2015-11-03
Adds a queue_maxsize param to Celery() that only works for redis brokers.

halfak:
Note that this change will do a few things:
* Adds a new error response (503 -- server overloaded)
* A whole request will fail instantly when trying to score revisions if the queue is too full
* When the queue is too full, `score()` raises `errors.ScoreProcessorOverloaded`
* A WARNING will be raise when the queue is overfull
* A WARNING will be raised if queue_maxsize is set and Celery is not using a redis broker (because queue_maxsize can't be respected)
* Adds a `score_processor_overloaded` metric to the metrics_collectors.

### DIFF
--- a/ores/score_processors/celery.py
+++ b/ores/score_processors/celery.py
@@ -118,8 +118,11 @@ class Celery(Timeout):
         # properties
         if self.redis is not None:
             queue_size = self.redis.llen("celery")
-            logger.warning("Queue size is too full {0}".format(queue_size))
-            return queue_size > self.queue_maxsize
+            if queue_size > self.queue_maxsize:
+                logger.warning("Queue size is too full {0}".format(queue_size))
+                return True
+            else:
+                return False
         else:
             return False
 
--- a/ores/score_processors/celery.py
+++ b/ores/score_processors/celery.py
@@ -218,6 +218,9 @@ REDIS_URL_RE = re.compile(
     r"(:(?P<port>[0-9]+))?" +
     r"(/(?P<db>[0-9]+))?"
 )
+"""
+Extracts the components of a redis URL as groups.
+"""
 
 
 def redis_from_url(url):
--- a/ores/score_processors/celery.py
+++ b/ores/score_processors/celery.py
@@ -9,6 +9,7 @@ import redis
 import revscoring.errors
 from celery.signals import before_task_publish
 
+from .. import errors
 from ..metrics_collectors import MetricsCollector
 from ..score_caches import ScoreCache
 from ..util import jsonify_error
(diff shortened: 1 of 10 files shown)


ores/score_processors/celery.py
```diff
@@ -102,7 +112,21 @@ def _generate_id(self, context, model, rev_id):
 
         return ":".join(str(v) for v in [context, model, rev_id, version])
 
+    def _queue_full(self):
+        # Check redis to see if the queue is too big.  If not, lump it on!
+        # This will result in a race condition, but it should have OK
+        # properties
+        if self.redis is not None:
+            queue_size = self.redis.llen("celery")
```
yuvipanda: Can we get this name from celery itself? Can we set this name in celery?

ores/score_processors/celery.py
```diff
@@ -102,7 +112,21 @@ def _generate_id(self, context, model, rev_id):
...
+            logger.warning("Queue size is too full {0}".format(queue_size))
```
yuvipanda: This is logging a warning no matter what?

halfak: Ha.  Yeah.

ores/score_processors/celery.py
```diff
@@ -102,7 +112,21 @@ def _generate_id(self, context, model, rev_id):
...
+            return queue_size > self.queue_maxsize
+        else:
+            return False
+
     def _score(self, context, model, rev_ids, caches=None):
+        if self._queue_full():
+            raise queue.Full()
```
yuvipanda: Is this just an exception? I wonder if we should have a specialized exception here for this instead, since queue.Full() probably has a specific meaning about the queue module etc than a generic queue system.... But if this makes sense to you I'm cool with it.

ores/score_processors/celery.py
```diff
@@ -174,10 +198,42 @@ def from_config(cls, config, name, section_key="score_processors"):
...
+                                   if k not in ('class', 'timeout',
+                                                'queue_maxsize')})
 
         return cls(scoring_contexts, application=application, timeout=timeout,
-                   score_cache=score_cache,
+                   score_cache=score_cache, queue_maxsize=queue_maxsize,
                    metrics_collector=metrics_collector)
+
+
+REDIS_URL_RE = re.compile(
```
yuvipanda: Document what this regex is for! Also can you not use the URI parser for this instead of regexes?

ores/score_processors/celery.py
halfak: I'm happy either way.  More specific might be nice. I'll give it a try.

ores/score_processors/celery.py
halfak: I've looked everywhere I could think of and I couldn't figure out how to get the name of the default queue.

ores/score_processors/celery.py
halfak: Seems like it might be more complicated to handle the uri parts.  

```
>>> parse.urlparse("redis://foobar@localhost/0")
ParseResult(scheme='redis', netloc='foobar@localhost', path='/0', params='', query='', fragment='')
```

ores/score_processors/celery.py
halfak: Now called errors.ScoreProcessorOverloaded

PR #107, opened 2015-12-09
Configures dev_server utility with rule-based scorer model.

halfak:
This will be useful in testing. 

RevIdScorer returns the last two digits in a rev_id as a prediction (378428499 = 99% & True).

### DIFF
--- a/ores/scorer_models/rev_id_scorer.py
+++ b/ores/scorer_models/rev_id_scorer.py
@@ -1,0 +1,51 @@
+from revscoring import ScorerModel
+from revscoring.datasources import revision
+from revscoring.features import Feature
+
+
+def process_last_two_in_rev_id(rev_id):
+    return int(str(rev_id)[-2:])
+
+last_two_in_rev_id = Feature(
+    "revision.last_two_in_rev_id",
+    process_last_two_in_rev_id,
+    returns=int,
+    depends_on=[revision.id]
+)
+
+
+class RevIdScorer(ScorerModel):
+    """
+    Implements a basic, testing scorer that predicts whether a revision ID's
+    last two digits are greater than 50.
+    """
+    def __init__(self, version=None):
+        super().__init__([last_two_in_rev_id], version=version)
+
+    def score(self, feature_values):
+        probability = feature_values[0] / 100
+
+        if probability > 0.5:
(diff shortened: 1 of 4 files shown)


ores/scorer_models/rev_id_scorer.py
```diff
@@ -0,0 +1,51 @@
...
+class RevIdScorer(ScorerModel):
+    """
+    Implements a basic, testing scorer that predicts whether a revision ID's
+    last two digits are greater than 50.
+    """
+    def __init__(self, version=None):
+        super().__init__([last_two_in_rev_id], version=version)
+
+    def score(self, feature_values):
+        probability = feature_values[0] / 100
```
Ladsgroup: one of them should be float otherwise it gives zero ( 1 / 10 = 0 but 1/10.0 = 0.1)

halfak: Only true in crazy python 2 land.  Join the python 3 revolution where divisions don't silently do stupid things. :)

Ladsgroup: :D

he7d3r:
Could you elaborate on how this will be used?

halfak:
Yes!  So, we're working on testing the new ORES extension, but one problem we have is setting up a scoring model that can work against the vagrant wiki or testwiki or whatever!  It would be a pain if we set up a machine learning model specifically for test wikis.  So instead, we set up a rule-based scorer model that requires no assets to function and returns a deterministic prediction.  This will help us test systems that *use ORES* independent of the models the useful models that ORES hosts.  

In the short term, we want to get the testwiki ORES server included as a role for MediaWiki vagrant and to get a testwiki model loaded onto ores.wmflabs.org for testing.  Both of those will use this scorer model.

halfak:
Relevant tasks:
* https://phabricator.wikimedia.org/T120956 -- ORES test server for vagrant/testwiki (this work)
* https://phabricator.wikimedia.org/T118039 -- Live test the ORES extension (the reason we're doing it)
* https://phabricator.wikimedia.org/T112856 -- Write prototype ORES extension (the main task we're working on)

he7d3r:
Thanks!

PR #108, opened 2015-12-09
Removes revscoring as a submodule.

halfak:
To be added to https://github.com/wiki-ai/ores-wikimedia-config


halfak:
See https://github.com/wiki-ai/ores-wikimedia-config/pull/42

PR #114, opened 2016-01-12
Minor things

arlolra:
Just some things I ran into setting up my dev env.

### DIFF
--- a/ores/ores.py
+++ b/ores/ores.py
@@ -6,10 +6,7 @@ This script provides access to a set of utilities for ORES
 * celery_worker -- Starts a "ScoreProcessor" celery worker
 * precached -- Starts a daemon that requests scores for revisions as they happen
 
-Usage:
-    ores (-h | --help)
-    ores <utility> [-h | --help]
-
+%(usage)s
 Options:
     -h | --help  Shows this documentation
     <utility>    The name of the utility to run
@@ -20,8 +17,9 @@ from importlib import import_module
 
 
 USAGE = """Usage:
-    ores (-h | --help)
-    ores <utility> [-h | --help]\n"""
+    %(progname)s (-h | --help)
+    %(progname)s <utility> [-h | --help]
+""" % { "progname": sys.argv[0] }
 
 
 def main():
@@ -30,7 +28,7 @@ def main():
         sys.stderr.write(USAGE)
         sys.exit(1)
     elif sys.argv[1] in ("-h", "--help"):
-        sys.stderr.write(__doc__ + "\n")
+        sys.stderr.write(__doc__ % { "usage": USAGE })
         sys.exit(1)
     elif sys.argv[1][:1] == "-":
         sys.stderr.write(USAGE)


arlolra:
`ImportError: No module named 'statsd'`
`ImportError: No module named 'pylru'`

These errors seem unrelated ... let's see what updating `.travis.yml` does.

requirements.txt
```diff
@@ -8,3 +8,4 @@ socketIO-client == 0.5.6
 stopit >= 1.1.1, < 1.1.999
 yamlconf >= 0.0.3, < 0.0.999
 revscoring >= 0.7.10
+redis
```
halfak: Redis is an optional dependency.  You can run with an in-memory LRU (pylru) or no cache at all.

arlolra: Did you see the commit message?

> Not sure if this is a hard dependency, but the test config seems to want it.

So is the correct patch to set the default config to no cache at all?

ores/ores.py
```diff
@@ -30,7 +28,7 @@ def main():
         sys.stderr.write(USAGE)
         sys.exit(1)
     elif sys.argv[1] in ("-h", "--help"):
-        sys.stderr.write(__doc__ + "\n")
+        sys.stderr.write(__doc__ % { "usage": USAGE })
```
halfak: Makes it a bit harder to read, but I appreciate that it reduces code duplication.  I wish there was a better way to do this.

arlolra: My main concern here was distinguishing `./utility` from `ores`.

requirements.txt
halfak: I don't see where the test config wants it.  Were you unable to run the test server without it?

ores/ores.py
halfak: Hmm.  It seems like we shouldn't make this distinction.  `./utility` is intended to be run under the dev environment and operate as closely as to the `ores` utility as possible.  Maybe you are hoping to avoid user-error somehow?

arlolra: Yes, user error.

I ran `./utility -h`. It told me to run `ores dev_server`. So I did `./utility ores dev_server` :)

requirements.txt
arlolra: > Were you unable to run the test server without it?

Nope. The error was,

```
λ (master) ./utility dev_server
Traceback (most recent call last):
  File "./utility", line 4, in <module>
    ores.main()
  File "/Users/arlolra/Work/Wikimedia/ores/ores/ores.py", line 48, in main
    module.main(sys.argv[2:])
  File "/Users/arlolra/Work/Wikimedia/ores/ores/utilities/dev_server.py", line 40, in main
    app = server.configure(config)
  File "/Users/arlolra/Work/Wikimedia/ores/ores/wsgi/server.py", line 6, in configure
    from ..score_processors import ScoreProcessor
  File "/Users/arlolra/Work/Wikimedia/ores/ores/score_processors/__init__.py", line 3, in <module>
    from .celery import Celery
  File "/Users/arlolra/Work/Wikimedia/ores/ores/score_processors/celery.py", line 7, in <module>
    import redis
ImportError: No module named 'redis'
```

halfak: Made it optional again in https://github.com/wiki-ai/ores/commit/5d135f848d0c3f084a7ffe3d67249a6d57f18ca2

ores/ores.py
halfak: Hmmm...  That's a good point.  This is a good change.

Issue #101, opened 2015-10-19
Provide datasource/feature overrides for scoring


ragesoss:
As a follow-on to #100, it would be great to be able to send simply a set of feature scores (for example, the current features of a real revision, but with 10 extra references) and get the model prediction for a revision with those features.

This would let us get a better feel for the behavior of the model, and also let us experiment with using the model to suggest specific interventions that will most improve a page.

halfak:
I just talked to @arlolra about getting this work done.  We discussed the steps that I thought needed to be taken to implement this in ORES.  Here's a quick summary:

Update ores/wsgi/routes/scores.py to handle parameters whose strings match datasource names (e.g. "datasource.revision.text") 
https://github.com/wiki-ai/ores/blob/master/ores/wsgi/routes/scores.py#L109

Update the basic score_processor so that it does not cache scores generated when datasources/features are provided as params. 
https://github.com/wiki-ai/ores/blob/master/ores/score_processors/score_processor.py#L164

We also talked about how we'd encode types in the datasources/features.  I think that JSON will work nicely but will look a bit ugly for strings.  Parameter names will need to be prefixes by "datasource." or "feature." in order to match the datasources and features of interest and that will provide us a ~namespace to make sure we don't overlap with other params (e.g. "precached")

halfak:
OK.  Proposal time!  I'm way overdue here.  

So, recent changes to revscoring (in the jump to `1.0.x`) have made all `revscoring.Feature` and `revscoring.Datasource` identifiable by a unique string.  This would allow us to provide key-value pairs where *key* is the unique string that identifies the `Datasource`/`Feature` and *value* is the value to place in the cache when extracting features and scoring.  Most of the potential values in *value* will be JSON-able (soon, all of them https://github.com/wiki-ai/revscoring/issues/212), so I think we should use JSON encoding for them.

I see two good options for how we could pass these key-value caches to ORES: url-ended params & a flat JSON blob.  To demonstrate the differences between these options, I will demonstrate with a request to get the article quality `wp10` prediction for `rev_id=707605082`, the last edit to the article on [Biology](https://en.wikipedia.org/w/index.php?title=Biology&oldid=707605082).  In this case, we'll hard-code the number of level-2 headers to 10 and set the text to be "bar". 

## Url-encoded params

* GET Request: `/scores/enwiki/wp10/707605082?feature.wikitext.revision.lvl_2_headers=10&datasource.revision.text="bar"`
[...]

PR #115, opened 2016-01-12
Parse args for features / datasources of interest

arlolra:
A start on #101

Some questions inline.

### DIFF
--- a/ores/scoring_contexts/scoring_context.py
+++ b/ores/scoring_contexts/scoring_context.py
@@ -70,7 +70,7 @@ class ScoringContext(dict):
         root_ds = [d for d in dependencies.dig(features)
                    if isinstance(d, Datasource)]
         error_root_vals = self.extractor.extract(rev_ids, root_ds,
-                                                 cache=caches)
+                                                 caches=caches)
         root_ds_caches = {}
         for rev_id, (error, root_vals) in zip(rev_ids, error_root_vals):
             if error is None:
--- a/ores/scoring_contexts/tests/test_scoring_context.py
+++ b/ores/scoring_contexts/tests/test_scoring_context.py
@@ -23,8 +23,8 @@ def test_scoring_context():
 
     FakeExtractor = namedtuple("Extractor", ['extract', 'solve', 'language'])
 
-    def fake_extract(rev_ids, dependents, cache=None):
-        caches = cache or defaultdict(dict)
+    def fake_extract(rev_ids, dependents, caches=None):
+        caches = caches or defaultdict(dict)
         for rev_id in rev_ids:
             cache = caches[rev_id]
             if rev_id % 5 != 0:
--- a/ores/wsgi/util.py
+++ b/ores/wsgi/util.py
@@ -2,6 +2,8 @@ from collections import defaultdict
 
 from flask.ext.jsonpify import jsonify
 
+import json
+
 
 class ParamError(Exception):
     pass
@@ -61,3 +63,18 @@ def format_output(context, scores, model_info, warning=None, notice=None):
         for rev_id in scores[model]:
             output['scores'][context][model]['scores'][rev_id] = scores[model][rev_id]
     return jsonify(output)
+
+
+def parse_features(request):
+    """Parse values for features / datasources of interest."""
+    caches = {}
+    try:
+        for k, v in request.values.items():
+            if k == "caches":
+                caches = json.loads(v)
+                break
+            elif k.startswith(("feature.", "datasource.")):
+                caches[k] = json.loads(v)
+    except Exception as e:
+        return e, None
+    return None, caches
--- a/ores/score_processors/score_processor.py
+++ b/ores/score_processors/score_processor.py
@@ -161,7 +161,8 @@ class SimpleScoreProcessor(ScoreProcessor):
                 try:
                     score = self._process(context, model, cache)
                     scores[rev_id] = score
-                    self._store(context, model, rev_id, score)
+                    if caches is None:
+                        self._store(context, model, rev_id, score)
                 except Exception as error:
                     scores[rev_id] = {'error': jsonify_error(error)}
 
--- a/ores/wsgi/routes/v1/scores.py
+++ b/ores/wsgi/routes/v1/scores.py
@@ -5,7 +5,7 @@ from flask.ext.jsonpify import jsonify
 
 from ... import responses
 from .... import errors
-from ...util import ParamError, read_bar_split_param
+from ...util import ParamError, read_bar_split_param, parse_features
 
 
 def configure(config, bp, score_processor):
@@ -108,8 +108,8 @@ def configure(config, bp, score_processor):
         return jsonify(scores)
 
     # /scores/enwiki/reverted/4567890
-    @bp.route("/scores/<context>/<model>/<int:rev_id>/", methods=["GET"])
-    @bp.route("/v1/scores/<context>/<model>/<int:rev_id>/", methods=["GET"])
+    @bp.route("/scores/<context>/<model>/<int:rev_id>/", methods=["GET", "POST"])
+    @bp.route("/v1/scores/<context>/<model>/<int:rev_id>/", methods=["GET", "POST"])
     def score_revision(context, model, rev_id):
 
         # Check to see if we have the context available in our score_processor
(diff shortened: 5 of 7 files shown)


ores/wsgi/routes/scores.py
```diff
@@ -113,19 +115,41 @@ def score_revision(context, model, rev_id):
...
-        precache = "precache" in request.args
-
-        # If the model exists, score revisions with it and return the result
         if model not in score_processor[context]:
             return responses.not_found("Model '{0}' not available for {1}."
                                        .format(model, context))
+
+        params = {}
+        if request.method == "POST":
+            data = request.get_json()  # Assume POSTs are JSON.
```
halfak: Why not just let POST data be x-www-form-urlencoded and use the same code to parse?

ores/wsgi/routes/scores.py
```diff
@@ -113,19 +115,41 @@ def score_revision(context, model, rev_id):
...
+                        depth = params
+                        path = k.split(".")
+                        last = len(path) - 1
+                        for i, key in enumerate(path):
+                            if i == last:
+                                depth[key] = json.loads(v)
+                            else:
+                                depth = depth.setdefault(key, {})
+            except:
+                return responses.bad_request("Unabled to parse params.")
```
halfak: Should probably include some detail about what failed.

ores/wsgi/routes/scores.py
arlolra: People like JSON? Not sure.

We could add a check for `request.headers.get("Content-Type") == "application/json"` and accept all three?

ores/wsgi/routes/scores.py
arlolra: Like,

```
except Exception as e:
    return responses.bad_request("Unabled to parse params: %s" % str(e))
```

halfak: Yeah.  I think something like that would be great.

ores/wsgi/routes/scores.py
halfak: Seems like a lot of complication.  The cache can't have a hierarchy to it.  It's only one-dimensional.  Since JSON object keys must be strings, I'm not sure if we get any benefit from JSON here.  We'd be implementing JSON-soap.

arlolra:
Whoops, didn't mean to close that.

ores/wsgi/routes/scores.py
```diff
@@ -113,19 +115,41 @@ def score_revision(context, model, rev_id):
...
+
+        # Is this right?  What's a precache?
+        precache = ("precache" in request.args) and (len(params) == 0)
+
+        # FIXME: Do something with `params`.
```
halfak: scores_processor.score(content, model, [rev_id], cache=params, precache=precache)

halfak:
Sorry for the delay here.  I'm working on a write-up of the different ways we can have users provide dependencies for feature extraction for feedback.  Should have an update this weekend.

arlolra:
Sure, no rush. Enjoy your weekend :)

halfak:
I got a write-up together of what I see as the two good options we have.  See https://github.com/wiki-ai/ores/issues/101#issuecomment-194488301

arlolra:
Thanks, taking a look.

ores/wsgi/routes/scores.py
```diff
@@ -113,19 +115,38 @@ def score_revision(context, model, rev_id):
...
-                                               precache=precache)
-            except errors.ScoreProcessorOverloaded:
-                return responses.server_overloaded()
 
-            return jsonify(scores)
+        caches = {}
+        # Parse values for features / datasources of interest.
+        try:
+            for k, v in request.values.items():
+                if k == "caches":
```
arlolra: In the discussion, you said `cache` but argument to `score()` is `caches`.

ores/wsgi/routes/scores.py
```diff
@@ -113,19 +115,38 @@ def score_revision(context, model, rev_id):
...
+                    caches = json.loads(v)
+                    break
+                elif k.startswith(("feature.", "datasource.")):
+                    depth = caches
+                    path = k.split(".")
+                    last = len(path) - 1
+                    for i, key in enumerate(path):
```
arlolra: The keys are flat, right? Like `caches["feature.123.456"]`. So this depth spelunking isn't really desired?

ores/wsgi/routes/scores.py
halfak: Ahh yes.  That's right.  I was imagining that we'd only support cached value injection on the `/<context>/<model>/<revid>/` route where `caches` would be a simple `{<revid>: { ... our flat JSON ... }}`.

ores/wsgi/routes/scores.py
halfak: That's right.  I think that keeping them flat makes sense since that's how revscoring works with them.

arlolra:
Updated for `v1` / `v2`.

ores/wsgi/util.py
```diff
@@ -61,3 +63,18 @@ def format_output(context, scores, model_info, warning=None, notice=None):
         for rev_id in scores[model]:
             output['scores'][context][model]['scores'][rev_id] = scores[model][rev_id]
     return jsonify(output)
+
+
+# Parse values for features / datasources of interest.
```
Ladsgroup: you should put documents in docstrings after function definition

Ladsgroup: PEP256

arlolra: Ok, but there are number of other PEP8 warnings in the library. Should they be fixed? Should CI not pass until they are?

Ladsgroup: I'll fix all of PEP8 issues in another PR, don't worry. It'll pass without this but I'm saying you're providing documentation, do it in the most proper way.

halfak: Good Q.  Don't worry about pep 8 issues on the rest of the file structure (unless you want to).  We can get those in a follow up PR.

arlolra:
This latest build failure seems unrelated.

arlolra:
Or not, nevermind, let me fix that.

arlolra:
Should be good to go now.

ores/scoring_contexts/scoring_context.py
```diff
@@ -70,7 +70,7 @@ def extract_roots(self, model, rev_ids, caches=None):
         root_ds = [d for d in dependencies.dig(features)
                    if isinstance(d, Datasource)]
         error_root_vals = self.extractor.extract(rev_ids, root_ds,
-                                                 cache=caches)
+                                                 caches=caches)
```
halfak: This doesn't match the behavior of revscoring

ores/scoring_contexts/tests/test_scoring_context.py
```diff
@@ -23,8 +23,8 @@ def test_scoring_context():
 
     FakeExtractor = namedtuple("Extractor", ['extract', 'solve', 'language'])
 
-    def fake_extract(rev_ids, dependents, cache=None):
-        caches = cache or defaultdict(dict)
+    def fake_extract(rev_ids, dependents, caches=None):
```
halfak: This doesn't match the behavior of revscoring

ores/scoring_contexts/scoring_context.py
arlolra: https://github.com/wiki-ai/revscoring/blob/a5039bd3e70ccb0ae333ca91845a38a049de2925/revscoring/extractors/extractor.py#L23

arlolra: https://github.com/wiki-ai/ores/blob/master/config/ores-testwiki.yaml#L54

halfak: OK fair point.  I'm just going to fix this in revscoring now.  Will have a PR shortly.

arlolra: https://github.com/wiki-ai/revscoring/pull/251

You'll also need to make a release and bump the requirements.txt here.

ores/scoring_contexts/tests/test_scoring_context.py
halfak: Now it does!

halfak:
This was merged, but I had to rebase on the way so it seems to have not been caught.  Thanks for your work,  @arlolra.  We'll be deploying this soon.  I'll let you know.

arlolra:
```
- caches=caches
+ caches={rev_id: cache},
```

I see. Thanks @halfak!

PR #116, opened 2016-01-12
Try to appease travis

arlolra:
This now passes: https://travis-ci.org/arlolra/ores/builds/101945348

But the logs say,
> nose.plugins.cover: ERROR: Coverage not available: unable to import coverage module


halfak:
Yeah.. Will need to find a better solution for these optional dependencies long-term.  this is pretty good for now.

PR #121, opened 2016-02-17
Api v2

Ladsgroup:
It is mentioned in https://phabricator.wikimedia.org/T121054

### DIFF
--- a/ores/wsgi/routes/v2.py
+++ b/ores/wsgi/routes/v2.py
@@ -41,6 +41,12 @@ def configure(config, bp, score_processor):
         except ParamError as e:
             return responses.bad_request(str(e))
 
+        # Get model info
+        try:
+            model_info_req = set(read_bar_split_param(request, "model_info", type=str))
+        except ParamError as e:
+            return responses.bad_request(str(e))
+
         # Check if all the models are available
         missing_models = models - score_processor[context].keys()
         if len(missing_models) > 0:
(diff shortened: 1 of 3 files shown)


ores/wsgi/routes/v2.py
```diff
@@ -0,0 +1,179 @@
...
+    def score_revision_v2(context, model, rev_id):
+
+        # Check to see if we have the context available in our score_processor
+        if context not in score_processor:
+            return responses.not_found("No scorers available for {0}"
+                                       .format(context))
+
+        precache = "precache" in request.args
+        info = score_processor[context][model].info()
+        model_info = {model: {'version': info['version']}}
```
halfak: model.version

ores/wsgi/routes/v2.py
```diff
@@ -0,0 +1,178 @@
...
+            rev_ids = set(read_bar_split_param(request, "revids", type=int))
+        except ParamError as e:
+            return responses.bad_request(str(e))
+
+        if len(rev_ids) == 0:
+            return responses.bad_request("No revids provided.")
+
+        precache = "precache" in request.args
+
+        # Generate scores for each model and merge them together
```
halfak: # Get model info for each model and merge them together

ores/wsgi/routes/v2.py
```diff
@@ -0,0 +1,178 @@
...
+                model_scores = score_processor.score(
+                    context, model, rev_ids, precache=precache)
+                scores[model] = model_scores
+        except errors.ScoreProcessorOverloaded:
+            return responses.server_overloaded()
+        return format_output(context, scores, model_info)
+
+    # /v2/scores/enwiki/reverted/?revids=456789|4567890
+    @bp.route("/v2/scores/<context>/<model>/", methods=["GET"])
+    def score_revisions_v2(context, model):
```
halfak: Should have score_revisions_v2 and score_revision_v2 share code.  I imagine a function that takes a `context`, `models` and `rev_ids` and produces the same structure dictionary.

ores/wsgi/routes/v2.py
```diff
@@ -0,0 +1,178 @@
...
+        # Get model info
+        try:
+            model_info_req = set(read_bar_split_param(
+                request, "model_info", type=str))
+        except ParamError as e:
+            return responses.bad_request(str(e))
+        if model_info_req:
+            for req in model_info_req:
+                try:
+                    model_info[model][req] = model_object.info()[req]
```
halfak: This will potentially be problematic. 

So, model_info.roc.fpr, model_info.roc.tpr, and model_info.roc.thresholds contains a set of values for plotting the actual ROC curve that a user would usually not want.  We need to find some way to make sure that we just return model_info.roc.auc by default when "roc" appears in "model_info" as well as some way to request all of the values for plotting when necessary.

halfak: On second though, let's merge this and I'll implement TestStatistic.format(stats, format="json") and make an update here in another PR.

PR #123, opened 2016-02-21
Redirect ORES main page to ORES page at meta

Ladsgroup:


halfak:
How about this? https://gist.github.com/halfak/fa632e01f5a6353fc18b

I think that's a better option than the redirect.

PR #130, opened 2016-03-15
Updates for revscoring 1.0.

halfak:


ores/scoring_contexts/scoring_context.py
```diff
@@ -63,15 +63,14 @@ def extract_roots(self, model, rev_ids, caches=None):
...
-                :class:`~revscoring.scorer_models.scorer_model.ScorerModel` to
+                The name of a :class:`~revscoring.ScorerModel` to
                 extract the roots for
         """
         features = self[model].features
         root_ds = [d for d in dependencies.dig(features)
                    if isinstance(d, Datasource)]
         error_root_vals = self.extractor.extract(rev_ids, root_ds,
-                                                 caches=caches)
+                                                 cache=caches)
```
arlolra: That doesn't seem right.

halfak: You're right.  It's weird.  This is due to a new overloading pattern in `Extractor.extract()` where, you can pass a single rev_id in rather than a list of rev_ids.  At some crazy moment, I decided that we should then make `cache` signular.   Rather than refactor and fix, I decided to just work with it for now and leave fixing this to a future refactor.  FWIW, it work as expected now.

PR #131, opened 2016-03-20
Swagger

halfak:


halfak:
To see swagger-ui working, start the server with `./utility dev_server` and point your browser to `http://localhost:8080/v1/` and `http://localhost:8080/v2/`.  To see the spec, append `spec` onto the past two URLs.

PR #136, opened 2016-03-28
Updates to v2 docs.

halfak:
Note that as I was reviewing the param names, I found that the 'cache' param was very confusing next to 'precache' since they mean very different things.  So I renamed 'cache' to 'inject' since providing data to the parameter performs a "dependency injection" during feature extraction.


halfak:
@arlolra, please review.  This changes the name of the `cache` param to `inject`

ores/wsgi/routes/v2/scores.py
```diff
@@ -6,7 +6,7 @@
 from ... import responses
 from .... import errors
 from ...util import (CacheParsingError, ParamError, format_output,
-                     parse_caches, read_bar_split_param)
+                     parse_injection, read_bar_split_param)
```
arlolra: Need to make this change for v1 as well, no?

ores/wsgi/routes/v2/scores.py
```diff
@@ -169,7 +195,7 @@ def score_revision_v2(context, model, rev_id):
             scores_doc[context][model]['info'] = model_info_doc
 
         try:
-            caches = parse_caches(request, rev_id)
+            caches = parse_injection(request, rev_id)
```
arlolra: Same

arlolra:
Sounds good to me, but v1 changes are missing. Unless the deprecated and removed already?

halfak:
Yeah.  Was thinking that this might be a nice incentive to move over to /v2/ and it would be nice not to have to double-maintain this in /v1/

arlolra:
Good idear. You'll want to clean up the obsolete code there though, and maybe inline `parse_injection`.

halfak:
I don't see any obsolete code in /v1/.  Not sure what you mean by "inline `parse_injection`".

arlolra:
Oh, sorry, didn't realize you removed `parse_caches` from v1 in efe0b311. Ignore.

re: inline, I mean, since there's only one call to `parse_injection`, it doesn't really need to be a utility anymore.

halfak:
Oh! I see.  I kind of like having a set of param parsing utilities in util.py for now.  We should consider refactoring, if necessary, in a future PR.

arlolra:
Ok, LGTM

PR #141, opened 2016-05-07
Error to logger when scoring is errored

Ladsgroup:


ores/score_processors/score_processor.py
```diff
@@ -59,7 +59,7 @@ def _get_root_ds(self, context, model, rev_ids, caches=None):
         start = time.time()
         roots = scoring_context.extract_roots(model, rev_ids, caches=caches)
         duration = time.time() - start
-        logger.debug("Extracted root datasources for " +
+        logger.debug("Extracted root datasources for "
```
halfak: Is not using an explicit "+" for concatenation not recommended?  I've been looking for something that would tell us about a best practice here and I can't find anything.

PR #143, opened 2016-05-22
Send watchdog signal to systemd for precaching

Ladsgroup:


halfak:
Only real concern is the GPL on the code.  It could break our MIT.  Who could we have comment on the Right(TM) way to do this?

halfak:
Looks good.  Were you able to run a test and demonstrate that systemd can bring the service back up?

Ladsgroup:
I tested, it has a minor issue and now works like a charm.

PR #147, opened 2016-06-14
Adds celery and wsgi applications.

halfak:
This replaces the functionality that was previously captured in https://github.com/wiki-ai/ores-wikimedia-config

You can run the application-based services with the regular utility: `./utility application.wsgi` or `./utility application.celery`.

Otherwise, you can configure uwsgi with `"ores.applications.wsgi:build()"`

### DIFF
--- a/ores/applications/celery.py
+++ b/ores/applications/celery.py
@@ -1,0 +1,39 @@
+"""
+Runs a celery worker.
+
+Usage:
+    celery [--config-dir=<path>]... [--logging-config=<path>]
+
+Options:
+    -h --help                Prints this documentation
+    --config-dir=<path>      The path to a directory containing configuration
+                             [default: config/]
+    --logging-config=<path>  The path to a logging configuration file
+"""
+
+import logging
+
+import docopt
+
+from ..score_processors import Celery
+from .util import build_config
+
+
+def main(argv=None):
+    args = docopt.docopt(__doc__, argv=argv)
+
+    run(config_dirs=args['--config-dir'],
+        logging_config=args['--logging-config'])
+
+
+def run(*args, **kwargs):
+    application = build(*args, **kwargs)
+    logging.getLogger('ores').setLevel(logging.DEBUG)
+    application.worker_main(argv=["celery_worker", "--loglevel=INFO"])
+
+
+def build(*args, **kwargs):
+    config = build_config(*args, **kwargs)
+    score_processor = Celery.from_config(
+        config, config['ores']['score_processor'])
+    return score_processor.application
--- a/ores/applications/util.py
+++ b/ores/applications/util.py
@@ -1,0 +1,27 @@
+import glob
+import os.path
+import yamlconf
+import logging.config
+
+DEFAULT_DIRS = ["config/", "/etc/ores/"]
+
+
+def build_config(config_dirs=DEFAULT_DIRS,
+                 logging_config="logging_config.yaml"):
+    config_paths = []
+    for directory in config_dirs:
+        config_paths.extend(glob.glob(os.path.join(directory, "*.yaml")))
+    config_paths.sort()
+    config = yamlconf.load(*(open(p) for p in config_paths))
+
+    if logging_config is not None:
+        with open(logging_config) as f:
+            logging_config = yamlconf.load(f)
+            logging.config.dictConfig(logging_config)
+
+    if 'data_paths' in config['ores'] and \
+       'nltk' in config['ores']['data_paths']:
+        import nltk
+        nltk.data.path.append(config['ores']['data_paths']['nltk'])
+
+    return config
--- a/ores/applications/util.py
+++ b/ores/applications/util.py
@@ -8,17 +8,22 @@ DEFAULT_DIRS = ["config/", "/etc/ores/"]
 
 def build_config(config_dirs=DEFAULT_DIRS,
                  logging_config="logging_config.yaml"):
-    config_paths = []
+    # Loads files in alphabetical order based on the bare filename
+    config_file_paths = []
     for directory in config_dirs:
-        config_paths.extend(glob.glob(os.path.join(directory, "*.yaml")))
-    config_paths.sort()
-    config = yamlconf.load(*(open(p) for p in config_paths))
+        dir_glob = os.path.join(directory, "*.yaml")
+        config_file_paths.extend((os.path.basename(path), path)
+                                 for path in glob.glob(dir_glob))
+    config_file_paths.sort()
+    config = yamlconf.load(*(open(p) for fn, p in config_file_paths))
 
+    # Load logging config if specified
     if logging_config is not None:
         with open(logging_config) as f:
             logging_config = yamlconf.load(f)
             logging.config.dictConfig(logging_config)
 
+    # Add nltk data path if specified.
     if 'data_paths' in config['ores'] and \
        'nltk' in config['ores']['data_paths']:
         import nltk
--- a/ores/applications/wsgi.py
+++ b/ores/applications/wsgi.py
@@ -1,0 +1,36 @@
+"""
+Runs a uwsgi server.
+
+Usage:
+    wsgi [--config-dir=<path>]... [--logging-config=<path>]
+
+Options:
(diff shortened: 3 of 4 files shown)


ores/applications/celery.py
```diff
@@ -0,0 +1,58 @@
...
+
+from ..score_processors import Celery
+
+DEFAULT_DIRS = ["config/", "/etc/ores/"]
+
+
+def main(argv=None):
+    args = docopt.docopt(__doc__, argv=argv)
+
+    run(config_dirs=args['--config-dir'],
```
schana: Is it possible for the arg to be multiple config directories? Either way, it looks like `build()` is expecting a list.

ores/applications/celery.py
```diff
@@ -0,0 +1,58 @@
...
+
+def run(*args, **kwargs):
+    application = build(*args, **kwargs)
+    logging.getLogger('ores').setLevel(logging.DEBUG)
+    application.worker_main(argv=["celery_worker", "--loglevel=INFO"])
+
+
+def build(config_dirs=DEFAULT_DIRS, logging_config="logging_config.yaml"):
+    config_paths = []
+    for dir in config_dirs:
```
schana: `dir` is a built-in function - a different name would be better

ores/applications/celery.py
```diff
@@ -0,0 +1,58 @@
...
+        config_paths.extend(glob.glob(os.path.join(dir, "*.yaml")))
+    config_paths.sort()
```
schana: How is the naming of that yaml files supposed to work? If `config_paths` has `/a/b/c/99-foo.yaml` and `/z/y/x/00-bar.yaml`, would the loading be in the expected order?

ores/applications/celery.py
```diff
@@ -0,0 +1,58 @@
...
+
+def main(argv=None):
+    args = docopt.docopt(__doc__, argv=argv)
+
+    run(config_dirs=args['--config-dir'],
+        logging_config=args['--logging-config'])
+
+
+def run(*args, **kwargs):
+    application = build(*args, **kwargs)
```
schana: Is this called from anywhere else? Why are the `*args, **kwargs` arguments needed as opposed to explicit ones?

ores/applications/wsgi.py
```diff
@@ -0,0 +1,56 @@
+"""
```
schana: This seems to have a lot of the same code as `celery.py`. Could we pull out the similarities?

ores/applications/celery.py
halfak: This will return a list, yes.

ores/applications/celery.py
halfak: Yes.  This is the intended pattern.

ores/applications/celery.py
halfak: In the utilities in the project, we split param parsing and cleanup from actually running a script.  This is the separation of responsibilities between main() and run().  Generally, it's best to pass *args, **kwargs rather than explicit arguments.  If run() needs an argument that build() can't take, that can still be handled semi-explicitly.  

E.g. 
```
def run(*args, runs_new_arg=None, **kwargs):
    ...
```

ores/applications/wsgi.py
halfak: Moved to util.py

halfak:
I just confirmed that the module pattern with `build()` works for celery too

```
$ celery worker -A "ores.applications.celery:build()"
```

halfak:
Please recheck @schana

ores/applications/celery.py
schana: Ah, it seems the `...` denotes a list. Consider me not a fan of `docopt`.

ores/applications/celery.py
schana: If that's the case, then I would suggest naming the methods such that their intent is clear.

> Generally, it's best to pass args, *kwargs rather than explicit arguments.
Source? This is the opposite of what I've encountered elsewhere.

ores/applications/util.py
```diff
@@ -0,0 +1,27 @@
...
+    config_paths.sort()
+    config = yamlconf.load(*(open(p) for p in config_paths))
+
+    if logging_config is not None:
+        with open(logging_config) as f:
+            logging_config = yamlconf.load(f)
+            logging.config.dictConfig(logging_config)
+
+    if 'data_paths' in config['ores'] and \
+       'nltk' in config['ores']['data_paths']:
```
schana: This would probably be more clear with exception based control flow. Also, it seems fairly common throughout ores to have dynamic imports based on configuration. Is there enough variation in configurations to warrant that?

ores/applications/util.py
```diff
@@ -0,0 +1,27 @@
+import glob
+import os.path
+import yamlconf
+import logging.config
+
+DEFAULT_DIRS = ["config/", "/etc/ores/"]
```
schana: Why specify these here when they're also in the docopt stuff?

ores/applications/wsgi.py
```diff
@@ -0,0 +1,36 @@
...
+def main(argv=None):
+    args = docopt.docopt(__doc__, argv=argv)
+    run(config_dirs=args['--config-dir'],
+        logging_config=args['--logging-config'])
+
+
+def run(*args, **kwargs):
+    application = build(*args, **kwargs)
+    logging.getLogger('ores').setLevel(logging.DEBUG)
+    application.debug = True
```
schana: Should `DEBUG` be a configuration value and not hard-coded?

ores/ores.py
```diff
@@ -6,6 +6,11 @@
 * celery_worker -- Starts a "ScoreProcessor" celery worker
 * precached -- Starts a daemon that requests scores for revisions as they happen
 
+You can also launch a set of production like applications
+
+* applications.wsgi -- A wsgi server
+* applications.celery -- A celery worker
+
```
schana: Why not have different entry points be specified in the setup.py, instead of maintaining this separate file?

ores/applications/util.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def build_config(config_dirs=DEFAULT_DIRS,
+                 logging_config="logging_config.yaml"):
+    config_paths = []
+    for directory in config_dirs:
+        config_paths.extend(glob.glob(os.path.join(directory, "*.yaml")))
+    config_paths.sort()
```
Ladsgroup: This would cause tons of issues for prod. you need to sort them in glob.glob(os.path.join(directory, "*.yaml")) and then extend.

Ladsgroup: That was the reason behind a 20 min. downtime today

ores/applications/wsgi.py
schana: Can we have separate commits and then squash when merging? It makes it easier to track what happens through the life of a pull request.

ores/applications/util.py
Ladsgroup: See this fix: https://gerrit.wikimedia.org/r/#/c/295214/

ores/applications/celery.py
halfak: Duly noted

ores/applications/wsgi.py
halfak: That's not how we usually operate.  I'm not sure I see the value.

ores/applications/util.py
halfak: Because, generally, these applications will be configured via uwsgi or celery worker and will not be called as command-line utilities.  See my notes on the initial PR.

ores/applications/wsgi.py
halfak: Quoting from another comment above:
> these applications will be configured via uwsgi or celery worker and will not be called as command-line utilities. See my notes on the initial PR.

The logging *is* configurable, but when executing the application script directly increased logging is very useful as this is only done when debugging.

ores/ores.py
halfak: A decision made long ago. See all of the rest of the utilities in this package.  If you'd like to discuss this pattern, it seems that it would be easier to start a whole separate thread.

PR #149, opened 2016-06-17
Wheelify assets

halfak:
* Uses new package `flask-swaggerui` to load Swagger-UI assets
* Uses new package `flask-wikimediaui` to load Wikimedia-UI assets
* Includes JQuery, JQuery-Tablesort and icon.js libs

See:
* https://github.com/wiki-ai/flask-wikimediaui ([pypi](https://pypi.python.org/pypi/flask-wikimediaui))
* https://github.com/wiki-ai/flask-swaggerui  ([pypi](https://pypi.python.org/pypi/flask-swaggerui))


halfak:
See also https://github.com/wiki-ai/flask-wikimediaui and https://github.com/wiki-ai/flask-swaggerui as well as https://pypi.python.org/pypi/flask-wikimediaui and https://pypi.python.org/pypi/flask-swaggerui.

PR #150, opened 2016-06-21
Adds nonbot_edit event to precached

halfak:


ores/utilities/precached.py
```diff
@@ -113,11 +113,11 @@ def on_change(self, change):
                     wikidb = change['wiki']
                     rev_id = change['revision']['new']
                     for model in score_on[('edit', wikidb)]:
-                        start = time.time()
                         executor.submit(get_score, wikidb, model, rev_id)
-                        logger.debug("GET {0} started in {1} seconds."
```
Ladsgroup: Make a comment here why you are removing debug, it would be nice to have it :)

halfak: OH yeah!  Good Q.  So, I realized that this debug line was really just noise in my testing.  It doesn't really serve a purpose other than that.  

FWIW, at one point I used it to make sure that requests were being fired off concurrently -- to make sure that the async pattern was working.

PR #152, opened 2016-06-26
Use strict_slashes = False

Ladsgroup:


Ladsgroup:
Some useful links:
http://stackoverflow.com/questions/17285826/flask-redirecting-multiple-routes
http://stackoverflow.com/questions/21050320/flask-301-response
http://permalink.gmane.org/gmane.comp.web.flask.general/5886

halfak:
Awesome!  This was such a pain and I was unable to find this solution in the past.

PR #159, opened 2016-07-15
Configurable home.html footer

halfak:


ores/wsgi/routes/__init__.py
```diff
@@ -14,7 +14,8 @@ def configure(config, bp, score_processor):
 
     @bp.route("/", methods=["GET"])
     def index():
-        return render_template("home.html")
+        return render_template(
+            "home.html", **config['ores'].get('home', {}))
```
Ladsgroup: Where is config['ores']['home']? Please add it in default settings.

PR #163, opened 2016-08-09
Adds metrics for precached.

halfak:


halfak:
Self-merging to test on staging.

PR #164, opened 2016-08-11
Fixes bug in cache storage pattern and with minor improvements to logging

halfak:


halfak:
This is the critical fix: https://github.com/wiki-ai/ores/pull/164/files#diff-5bac26b919014058e3e84feda6c2ab8aL224

PR #165, opened 2016-08-29
No cache headers for model info and scores

halfak:


Ladsgroup:
@akosiaris, Hey can you check this?

ores/wsgi/util.py
```diff
@@ -82,3 +84,17 @@ def parse_injection(request, rev_id):
...
     except Exception as e:
         raise CacheParsingError(e)
+
+
+def nocache(route):
+    @wraps(route)
+    def no_cache(*args, **kwargs):
+        response = make_response(route(*args, **kwargs))
+        response.headers['Cache-Control'] = \
+            "no-store, no-cache, must-revalidate, post-check=0, " + \
```
akosiaris: must-revalidate should be sent by servers IF AND ONLY IF failure to revalidate a request could result in incorrect operation (like a transaction failing), which is arguably not the case here.

ores/wsgi/util.py
```diff
@@ -82,3 +84,17 @@ def parse_injection(request, rev_id):
...
+            "pre-check=0, max-age=0"
```
akosiaris: The post-check and pre-check parts are IE specific and are generally frowned upon these days. And for the use case in mind possibly harmful. See  https://blogs.msdn.microsoft.com/ieinternals/2009/07/20/internet-explorers-cache-control-extensions/ where it is said

> If you want to prevent caching, do not include the post-check and pre-check directives. Doing so is completely unnecessary and results in wasted bandwidth and HTTP header processing cycles.

ores/wsgi/util.py
```diff
@@ -82,3 +84,17 @@ def parse_injection(request, rev_id):
...
+        response.headers['Pragma'] = 'no-cache'
```
akosiaris: this is redundant these days. Feel free to keep it in just to satisfy really really old clients but I doubt these clients would be accessing ORES (and even more having caching issues)

ores/wsgi/util.py
```diff
@@ -82,3 +84,17 @@ def parse_injection(request, rev_id):
...
+        response.headers['Expires'] = '-1'
```
akosiaris: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html says Expires needs to be in RFC 1123 days format. And -1 is not that. Either set it to epoch (Jan 1st 1970) or remove it

ores/wsgi/util.py
```diff
@@ -82,3 +84,16 @@ def parse_injection(request, rev_id):
...
+
+
+def nocache(route):
+    @wraps(route)
+    def no_cache(*args, **kwargs):
+        response = make_response(route(*args, **kwargs))
+        response.headers['Cache-Control'] = \
+            "no-store, no-cache, max-age=0"
+        response.headers['Pragma'] = 'no-cache'
+        response.headers['Expires'] = '01 Jan 1970' # Unix epoch
```
akosiaris: More like "Thu, 01 Jan 1970 00:00:00 GMT"

akosiaris:
Added one minor comment about the Date format, rest LGTM

halfak:
Self merging since receiving a LGTM and addressing the last note.  Thanks @akosiaris!

PR #172, opened 2016-10-19
Make it possible to run the tests from tox

adamwight:
Some people might think is convenient.  Tox will take care of the dependencies
and creates virtualenvs for testing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 minversion = 1.6
-envlist = flake8
+envlist = flake8,discover
 
 [flake8]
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -11,3 +11,9 @@ ignore = E126,E127
 commands = flake8
 deps = flake8
 basepython = python3.4
+
+[testenv:discover]
+commands = nosetests {posargs} {toxinidir}
+deps =
+    -rrequirements.txt
```
Ladsgroup: It looks very wrong at first glance but per https://media.readthedocs.org/pdf/tox/latest/tox.pdf it's correct. Super weird

PR #178, opened 2016-11-29
Two small additions for Installation section of README.

och:
Currently the guide assumes that the user has Python 3 in there system, which might not be the case for stable-ver. distros like Debian in which Python 2 is used by default.

I've also added the usage of the -h argument to encourage users to look at the help page.


Ladsgroup:
Thanks!

PR #180, opened 2016-11-30
Revert "Add celery.exceptions.TimeoutError to expected_errors in CeleryQueue"

halfak:
Reverts wiki-ai/ores#179

### DIFF
--- a/ores/scoring_systems/celery_queue.py
+++ b/ores/scoring_systems/celery_queue.py
@@ -57,8 +57,7 @@ class CeleryQueue(ScoringSystem):
                            revscoring.errors.DependencyError,
                            mwapi.errors.RequestError,
                            mwapi.errors.TimeoutError,
-                           errors.TimeoutError,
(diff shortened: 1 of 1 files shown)


halfak:
So, I don't think this is the right way to solve this problem.  I think we ought to handle the celery error within the sub-task

halfak:
It seems that we might want to handle the error here: https://github.com/wiki-ai/ores/blob/master/ores/scoring_systems/celery_queue.py#L92

Note that we *do* handle specific Timeout error and normalize them to our timeout error elsewhere.  
* https://github.com/wiki-ai/ores/blob/master/ores/scoring_systems/celery_queue.py#L151
* https://github.com/wiki-ai/ores/blob/master/ores/scoring_systems/process_pool.py#L44

halfak:
OK I've decided that this isn't worth reverting, but I will be submitting a follow-up PR to implement it the way I think is best.

PR #183, opened 2017-01-22
Minify json responses

Ladsgroup:


halfak:
Hey @Ladsgroup, per discussion in chat, I think we should return pretty-print by default (implicit `format=jsonpretty`) and use a `format=json` to return minified JSON responses.

halfak:
Thanks!

PR #185, opened 2017-02-10
Fixes and cleans up JSON prettyprint processing

halfak:
https://phabricator.wikimedia.org/T157721


Ladsgroup:
It's great but the PR contains much more than its title. It should be in separate PRs or be explained here. But since we have different commits, I'm okay.

PR #188, opened 2017-02-23
Add notice-like header

Ladsgroup:
It'll be something like https://dumps.wikimedia.org/


Ladsgroup:
https://phabricator.wikimedia.org/T150962

PR #189, opened 2017-03-01
Add precache endpoint to v2 API

Ladsgroup:
Bug: https://phabricator.wikimedia.org/T148714

### DIFF
--- a/ores/wsgi/responses.py
+++ b/ores/wsgi/responses.py
@@ -1,6 +1,9 @@
 from flask.ext.jsonpify import jsonify
 
 
+def no_content():
+    return "", 204
+
 def error(status, code, message):
     return jsonify({'error': {'code': code, 'message': message}}), status
 
--- a/ores/wsgi/routes/v2/__init__.py
+++ b/ores/wsgi/routes/v2/__init__.py
@@ -1,6 +1,7 @@
 from flask import request
 from flask_swaggerui import render_swaggerui
 
+from . import precache
 from . import scores
 from . import spec
 
@@ -14,6 +15,7 @@ def configure(config, bp, score_processor):
         else:
             return render_swaggerui(swagger_spec_path="/v2/spec/")
 
+    bp = precache.configure(config, bp, score_processor)
     bp = scores.configure(config, bp, score_processor)
     bp = spec.configure(config, bp, score_processor)
 
--- a/ores/wsgi/routes/v2/precache.py
+++ b/ores/wsgi/routes/v2/precache.py
@@ -1,0 +1,68 @@
+import traceback
+
+from flask import request
+from flask.ext.jsonpify import jsonify
+
+from ... import preprocessors, responses
+from .... import errors
+from ...util import ParamError, read_bar_split_param
+
+
+def configure(config, bp, scoring_system):
+
+    @bp.route("/v2/precache/", methods=["GET"])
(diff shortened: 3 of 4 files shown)


Pchelolo:
Cool! +1, although I don't know python :)

halfak:
What do you think about taking the whole rc blob rather than a set of query parameters?  That would give us the freedom to add new event types to our precaching config.  E.g. I want to add "new" as in "new page creation" some time soon for the draftquality model.

Pchelolo:
> What do you think about taking the whole rc blob rather than a set of query parameters? That would give us the freedom to add new event types to our precaching config. E.g. I want to add "new" as in "new page creation" some time soon for the draftquality model.

We can easily post the whole `revision-create` event unchanged. A bit of overhead in networking, but you'd be able to figure out whether it's a new page or just a new revision by looking at the `rev_parent_id` property. So no objections from our side.

Ladsgroup:
@Pchelolo How `revision-create` blob looks like? Can you point out to a documentation?

Pchelolo:
@Ladsgroup Here's the schema: https://github.com/wikimedia/mediawiki-event-schemas/blob/master/jsonschema/mediawiki/revision/create/1.yaml

Here's an example event:
```json
{
  "comment": "BLABLABLA",
  "database": "wikidatawiki",
  "meta": {
    "domain": "www.wikidata.org",
    "dt": "2017-03-03T21:41:21+00:00",
    "id": "248bdd8b-005a-11e7-804c-90b11c2fb49a",
    "request_id": "c6e37cf2-73cb-454d-95e7-32f70087cfee",
    "schema_uri": "mediawiki/revision/create/1",
    "topic": "mediawiki.revision-create",
    "uri": "https://www.wikidata.org/wiki/Q10022493"
  },
  "page_id": 11125321,
  "page_is_redirect": false,
  "page_namespace": 0,
  "page_title": "Q10022493",
  "performer": {
    "user_groups": [
      "bot",
      "*",
      "user",
      "autoconfirmed"
    ],
    "user_id": 123456,
    "user_is_bot": true,
    "user_text": "BLABLABLA"
  },
  "rev_content_format": "wikibase-item",
  "rev_content_model": "wikibase-item",
  "rev_id": 460857280,
  "rev_len": 6911,
  "rev_minor_edit": false,
  "rev_parent_id": 446118948,
  "rev_sha1": "28yefn57vzn4tooueol368xboox4jyy",
  "rev_timestamp": "2017-03-03T21:41:21Z"
}
```

Ladsgroup:
Added the data blob. This works in my localhost: http://0.0.0.0:8080/v2/precache/?data={%22database%22:%22testwiki%22,%20%22rev_id%22:%201234}. @Pchelolo: Please inform us if there is any change in the schema you mentioned.

Pchelolo:
@Ladsgroup Omg, I didn't keep track of this and didn't notice the latest changes! Oh no....

So, if you wanna get the whole `revision-create` event in the GET query parameter, that's pretty much impossible. The longest URI you can have is 2,083 characters, and since the event is pretty big and contains page titles (that can be really long in MediaWiki) the URI length can easily exceed the maximum.. So, this endpoint should be POST instead of a GET and provide the event in the body.

If you don't wanna have the whole event but just a sunset of it, I don't quite understand the move to the `?data` parameter, individual query params would've been better..

PR #191, opened 2017-03-23
Adds ScoreRequest and ScoreResponse and refactors.

halfak:


ores/score_request.py
```diff
@@ -0,0 +1,52 @@
...
+        self.model_names = set(model_names)
+        self.precache = precache
+        self.include_features = include_features
+        self.injection_caches = injection_caches or {}
+        self.model_info = model_info
+
+    def __str__(self):
+        return self.format()
+
+    def format(self, rev_id=None, model_name=None):
```
adamwight: This and __repr__ below would benefit from some docstrings.  What's the difference?  Why would I want to cast to string and how would I use the result?  If it's a magic format, maybe there should be more semantic accessors like "get_unique_key" or sumsuch.

ores/score_response.py
```diff
@@ -0,0 +1,37 @@
...
+        for rev_id, model_name, score in scores:
+            self.add_score(rev_id, model_name, score)
+        for rev_id, model_name, error in scores:
+            self.add_error(rev_id, model_name, error)
+        for rev_id, model_name, feature_vals in features:
+            self.add_features(rev_id, model_name, feature_vals)
+        for model_name, info_doc in model_info:
+            self.add_model_info(model_name, info_doc)
+
+    def add_score(self, rev_id, model_name, score):
```
adamwight: TODO in a later refactor: add_score(RevScoreObj(rev_id, model_name, score))

ores/scoring_context.py
```diff
@@ -34,18 +34,13 @@ def __init__(self, name, model_map, extractor):
...
-                             for field, value in formatted_info.items()
-                             if fields == "all" or field in fields}
-
-            model_info[model_name] = filtered_info
-
-        return model_info
+        formatted_info = self._get_model_info_for(model_name)
+        filtered_info = {field: value
+                         for field, value in formatted_info.items()
+                         if fields == [""] or field in fields}
```
adamwight: I'd like to see the `fields == [""]` test extracted into a is_all_fields predicate...  It looks arch-magical in there.

ores/scoring_systems/celery_queue.py
```diff
@@ -157,18 +147,16 @@ def _process_missing_scores(self, context_name, missing_model_set_revs,
 
         return rev_scores, score_errors
 
-    def _lookup_inprogress_results(self, context_name, model_names, rev_ids,
-                                   injection_caches=None, rev_scores=None):
-        context = self[context_name]
+    def _lookup_inprogress_results(self, request, response):
```
adamwight: /me perks up.  Things just got fancy!

ores/scoring_systems/scoring_system.py
```diff
@@ -5,10 +5,11 @@
 import stopit
 
 from .. import errors
-from ..errors import TimeoutError
+from ..errors import MissingContext, MissingModels, TimeoutError
 from ..metrics_collectors import Null
 from ..score_caches import Empty
-from ..util import jsonify_error, timeout
+from ..score_response import ScoreResponse
```
adamwight: Tangent: I see that Guido is currently recommending relative imports, but we should reconsider how we use them.  AIUI, the "spirit" of the law is that we are free to do whatever it takes to avoid having to refactor references later.  The way I interpret that, if the dependent and dependency files are related in a way that makes sense or might be pretty constant then a relative import makes sense, but in cases like this we should say "ores.errors" and "ores.score_response" cos those are the more permanent addresses.

I'm just picking through this myself, so no worries if you prefer the consistency of always using relative, don't want a bunch more churn, etc.

ores/wsgi/routes/__init__.py
```diff
@@ -6,6 +6,7 @@
 from . import ui
 from . import v1
 from . import v2
+from . import v3
```
adamwight: /me pops a peanut

ores/wsgi/util.py
```diff
@@ -35,23 +37,147 @@ def read_bar_split_param(request, param, default=None, type=str):
...
 
 
+def format_error(error):
+    error_type = error.__class__.__name__
+    message = str(error)
+
+    return {'error': {'type': error_type, 'message': message}}
+
+
+def build_score_request(scoring_system, request, context_name=None, rev_id=None,
```
adamwight: Could move this to score_request.build_from_web_request

I wouldn't mind a docstring here, to remind me of the parameter types.

ores/wsgi/util.py
```diff
@@ -35,23 +37,147 @@ def read_bar_split_param(request, param, default=None, type=str):
...
+        model_names = scoring_system[context_name].keys()
+
+    return ScoreRequest(context_name, rev_ids, model_names,
+                        precache=precache,
+                        include_features=include_features,
+                        injection_caches=injection_caches,
+                        model_info=model_info)
+
+
+def parse_rev_ids(request, rev_id):
```
adamwight: Should we do this in a new module, request_parse?  Fine to get there iteratively.

ores/wsgi/util.py
```diff
@@ -35,23 +37,147 @@ def read_bar_split_param(request, param, default=None, type=str):
...
+    if context_name not in precache_map:
+        return None
+
+    # Start building the response document
+    event_set = build_event_set(event)
+
+    model_names = {precache_map[context_name][e] for e in event_set
+                   if e in precache_map[context_name]}
+
+    if len(model_names) == 0:
```
adamwight: Is this worth raising an exception?

ores/wsgi/util.py
```diff
@@ -35,23 +37,147 @@ def read_bar_split_param(request, param, default=None, type=str):
...
+        return None
+
+    return ScoreRequest(context_name, model_names, rev_id)
+
+
+# TODO: This strategy for building up events is not sustainable.
```
adamwight: Would streaming be enough of an improvement?

halfak:
Here's some notes on what's going on here:

* ScoringSystem.score(ScoreRequest) --> ScoreResponse
  * This is a big change.  ScoringSystem.score() used to take a set of arguments (context_name, model_names, rev_ids, include_features=False, preache=False, model_info=None).  It would return a json-ready dict with a complex response structure. (https://github.com/wiki-ai/ores/blob/request_response/ores/scoring_systems/scoring_system.py#L36 )
  * The ScoreRequest wraps up all of those arguments and makes them easier to pass around (https://github.com/wiki-ai/ores/blob/request_response/ores/score_request.py )
  * The ScoreResponse wraps up all of the stuff that should appear in a json-ready response dictionary (https://github.com/wiki-ai/ores/blob/request_response/ores/score_response.py )
* WSGI Routes -- This made a few refactorings to wsgi routes nice and clean.
ScoreResponse formatters now handle all json-able dict creations
  * https://github.com/wiki- ai/ores/blob/request_response/ores/wsgi/routes/v1/util.py#L8
  * https://github.com/wiki-ai/ores/blob/request_response/ores/wsgi/routes/v2/util.py#L9
  * https://github.com/wiki-ai/ores/blob/request_response/ores/wsgi/routes/v3/util.py#L8
* ScoreRequest is generated from the wsgi (web request) context using one function (https://github.com/wiki-ai/ores/blob/request_response/ores/wsgi/util.py#L47 )
  * This function is nice because it can take some basic context information + a request object and build an appropriate ScoreRequest object.  I toyed with the idea of having each version of the API have its own function, but I soon realized that they all needed to solve the exact same problem.
* There are a few places this propagates to and that accounts for a huge amount of the change to lines-of-code
  * https://github.com/wiki-ai/ores/blob/request_response/ores/metrics_collectors/metrics_collector.py generally takes (request) rather than (context_name, models_names, ...) 
  * https://github.com/wiki-ai/ores/blob/request_response/ores/metrics_collectors/logger.py is affected by ^ 
  * https://github.com/wiki-ai/ores/blob/request_response/ores/metrics_collectors/statsd.py is affected by ^ 
* This has made a couple of things *way* easier
  * The https://github.com/wiki-ai/ores/tree/request_response/ores/wsgi/routes/v3 version of the API was *super easy* to put together
    * Note that Lots of lines of code come from the v3 swagger https://github.com/wiki-ai/ores/blob/request_response/ores/wsgi/templates/v3_swagger.json
  * Implementing a difference in cache hit/miss for precache requests was excessively trivial (https://github.com/wiki-ai/ores/blob/request_response/ores/metrics_collectors/statsd.py#L44 )
* Look at how short the score.py route files are now!  Two lines for most routes/functions!
  * https://github.com/wiki-ai/ores/blob/request_response/ores/wsgi/routes/v1/scores.py#L53
  * https://github.com/wiki-ai/ores/blob/request_response/ores/wsgi/routes/v2/scores.py#L42
  * https://github.com/wiki-ai/ores/blob/request_response/ores/wsgi/routes/v3/scores.py#L44

ores/wsgi/util.py
halfak: I like having it here because there's a few different ways we build requests.  E.g. from the command-line for running ORES offline.   

+1 for docstring

ores/wsgi/util.py
halfak: I like the `wsgi` module because it's purpose is to handle web stuff.

ores/wsgi/util.py
halfak: There's a difference between nothing-to-do and something-went-wrong.  In this case, return None means nothing-to-do.

ores/wsgi/util.py
halfak: Not exactly.  I think we build up a set of characteristics for each edit and a set of matching characteristics for each event.  In turn, we'll need to allow for configuration to require and exclude certain types of events.  e.g. `{'edit', '!bot', '!new'}` would only match non-page-creating non-bot edits.  `{'log', 'newuser', '!autocreate'}` would only match new user creations that create a new centralauth record.

halfak:
OK this thing is ready for final review.

PR #192, opened 2017-04-11
Switch to POST for precache endpoint in v3

Ladsgroup:
Bug: https://phabricator.wikimedia.org/T162627

### DIFF
--- a/ores/wsgi/routes/v3/precache.py
+++ b/ores/wsgi/routes/v3/precache.py
@@ -14,15 +14,15 @@ def configure(config, bp, scoring_system):
 
     precache_map = util.build_precache_map(config)
 
-    @bp.route("/v3/precache/", methods=["GET"])
+    @bp.route("/v3/precache/", methods=["POST"])
     @preprocessors.nocache
     @preprocessors.minifiable
     def precache_v3():
-        if 'event' not in request.args:
+        if 'event' not in request.form:
             return responses.bad_request(
                 "Must provide an 'event' parameter")
         try:
-            event = json.loads(unquote(request.args['event']).strip())
+            event = json.loads(unquote(request.form['event']).strip())
         except json.JSONDecodeError:
             return responses.bad_request(
                 "Can not parse event argument as JSON blob")
--- a/ores/wsgi/routes/v3/precache.py
+++ b/ores/wsgi/routes/v3/precache.py
@@ -1,6 +1,4 @@
-import json
 import logging
-from urllib.parse import unquote
 
 from flask import request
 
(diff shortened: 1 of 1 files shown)


ores/wsgi/routes/v3/precache.py
```diff
@@ -14,15 +14,15 @@ def configure(config, bp, scoring_system):
 
     precache_map = util.build_precache_map(config)
 
-    @bp.route("/v3/precache/", methods=["GET"])
+    @bp.route("/v3/precache/", methods=["POST"])
     @preprocessors.nocache
     @preprocessors.minifiable
     def precache_v3():
-        if 'event' not in request.args:
+        if 'event' not in request.form:
```
d00rman: I don't think we need this at all. The body can just be the event itself in JSON format.

ores/wsgi/routes/v3/precache.py
```diff
@@ -14,15 +14,15 @@ def configure(config, bp, scoring_system):
...
             return responses.bad_request(
                 "Must provide an 'event' parameter")
         try:
-            event = json.loads(unquote(request.args['event']).strip())
+            event = json.loads(unquote(request.form['event']).strip())
```
d00rman: We also need it to be able to support application/json POST bodies, not just encoded forms.

Pchelolo: I don't really think we even need to support form at all..

halfak: Agreed.  Let the whole post body be a JSON blob or allow for a get param (which is very useful in debugging)

Pchelolo: @halfak The `get` param is not really possible, since on average the URI-encoded revision-create even will exceed the maximum possible URI length ( 2,083 characters ), so let's stick with POST to avoid confusion?

ores/wsgi/routes/v3/precache.py
Ladsgroup: Done

ores/wsgi/routes/v3/precache.py
Ladsgroup: Done

halfak:
Hmm.  I'd like to be able to run tests from my browser.   I'm sure a sample event could stay below the 2k char limit for the purposes of testing.  I'd still like to have that get parameter.  I can't see how it would be that confusing.

Ladsgroup:
I use Postman (a chrome extension) to make POST request and debug. Besides having a fancy GUI, it has lots of features (such as prettifying json, known types, validations, etc.). Try it and you would love it.

halfak:
OK.  Fair enough :)

PR #197, opened 2017-04-27
Add scheme to config and use it in producing swagger spec

Ladsgroup:
Bug: https://phabricator.wikimedia.org/T162184


halfak:
Seems like a good enough solution to me.

PR #199, opened 2017-05-08
Rename default config file to play nicer with other files

tgr:
Config files are merged in alphabetical order with later files
overriding settings in earlier ones so apply the usual naming
convention <weight>-<name>. Also, drop the 'ores' part of the name
which does not say anything useful.


Ladsgroup:
Fine with me, let's wait for Aaron though.

halfak:
We usually use "00-main.yaml" as the primary config.  "100-testwiki.yaml" would sort after "00-main.yaml"  I proposed that we name the file "00-testwiki.yaml" or just use the common "00-main.yaml" name that we use in deployments.

tgr:
So maybe the fundamental problem is that we have a testwiki config that gets loaded whether you want it or nor (and cannot be easily disabled without dirtying your git repo). In which case it could be renamed to `testwiki.yaml.example` or something like that, or added to gitignore (although that's more hacky).

halfak:
> So maybe the fundamental problem is that we have a testwiki config that gets loaded whether you want it or not

Hmm... Not sure what you mean here.  It gets loaded if it is found in the expected default directory: `./config/`.  ORES can't start without a configuration file.  The testwiki config would not be loaded if you specify another configuration directory location.  

It'd like ORES to be *start-able* with a basic, test configuration out of the box.

halfak:
Done in https://github.com/wiki-ai/ores/commit/726dd13f5379624e0dd5dd4cbe7f41353d6f7595

PR #200, opened 2017-06-04
Migrate to EventStream in precaching

Ladsgroup:
Even though it's straightforward, I hadn't have time to test it.
Bug: https://phabricator.wikimedia.org/T166046

### DIFF
--- a/ores/utilities/precached.py
+++ b/ores/utilities/precached.py
@@ -20,17 +20,17 @@ If ran by systemd, it uses watchdog to stay alive.
                      [default: config]
     --verbose        Print debugging information
 """
-import concurrent.futures
+import json
 import glob
 import logging
 import os
 import sys
 import time
 from collections import defaultdict
+from sseclient import SSEClient as EventSource
 
 import docopt
 import requests
-import socketIO_client
 import yamlconf
 
 from ..metrics_collectors import MetricsCollector, Null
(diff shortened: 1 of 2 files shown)


ores/utilities/precached.py
```diff
@@ -66,142 +76,32 @@ def main(argv=None):
...
+            time.sleep(delay)
+        if notify:
+            watchdog_ping(*notify)
+        start = time.time()
+        response = requests.post(
+            ores_url + "/v3/precache/", data=json.dumps(change))
+
+        if response.status == 200:
+            logger.info("Scored {0} in {1} seconds."
+                        .format(json.dumps(change), round(time.time() - start, 3)))
```
Ladsgroup: I have a feeling that change will be a huge variable, we should only get rev id.

halfak: This is a fair point.  There may be some more important things.  E.g. 'bot', 'namespace', etc.   We should get that small subset or the whole thing.

halfak:
This needs a real world test before merging.

Ladsgroup:
Okay, I made lots of tests and made tons of fixes. It now works but server responds with "database" is not there (and "rev_id" is not there) because change schema of event stream is different than what is change propagation change scheme (ping @ottomata and @Pchelolo, Any plans to make them look alike?)

Also this needs to send user-agent.

ottomata:
Ya you are reading recent changes which is s different stream. We've been waiting for someone to need revision-create in the EventStreams public API before we added it in. Looks like this is it! :)

halfak:
Yeah!  please :)))

halfak:
See https://phabricator.wikimedia.org/T167670

halfak:
tested on my local machine.  I ran this command to set up precached daemon:

```
./utility precached https://stream.wikimedia.org/v2/stream/revision-create http://localhost:8080
```

PR #202, opened 2017-06-05
Add 404.html for general 404 errors

Ladsgroup:
Bug: https://phabricator.wikimedia.org/T113114

### DIFF
--- a/ores/wsgi/templates/404.html
+++ b/ores/wsgi/templates/404.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+<meta charset="utf-8">
+<title>Not Found</title>
+<link rel="shortcut icon" href="/favicon.ico">
+<style>
+* { margin: 0; padding: 0; }
+body { background: #fff; font: 15px/1.6 sans-serif; color: #333; }
+.content { margin: 7% auto 0; padding: 2em 1em 1em; max-width: 640px; }
+img { float: left; margin: 0 2em 2em 0; }
+a img { border: 0; }
+h1 { margin-top: 1em; font-size: 1.2em; }
+p { margin: 0.7em 0 1em 0; }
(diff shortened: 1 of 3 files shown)


ores/wsgi/templates/404.html
```diff
@@ -0,0 +1,26 @@
...
+a:hover { text-decoration: underline; }
+em { font-style: normal; color: #777; }
+</style>
+<div class="content" role="main">
+<a href="{{ host }}"><img src="{{ host }}/static/images/wmf.png" srcset="{{ host }}/static/images/wmf-2x.png 2x" alt="{{ alt }}" width=135 height=135></a>
+<h1>Page not found</h1>
+<p><em>{{ title }}</em></p>
+<p>We could not find the above page on our servers.</p>
+<div style="clear:both;"></div>
+<p>Alternatively, you can visit the <a href="/">Main Page</a> or read <a href="https://en.wikipedia.org/wiki/HTTP_404" title="Wikipedia: HTTP 404">more information</a> about this type of error.</p>
```
halfak: Should we direct people to English Wikipedia's main page?  Seems we should try to direct them to ORES main page.

Ladsgroup: I don't think we should direct people to English Wikipedia's main page. In other places it's the root path so for example https://fa.wikipedia.org/Example will direct people to Persian Wikipedia's main page. I think we should send them to ORES main page (which is the current state)

Pix1234:
Or maybe direct to a multi-lingul 404 page?

Ladsgroup:
@Pix1234 Nope, see the bug: https://phabricator.wikimedia.org/T113114 This is more in production cluster level.

ores/wsgi/templates/404.html
halfak: Oh!  i think I misread.

PR #206, opened 2017-06-09
Add attribution

melodykramer:
I added attribution, even though it's in the document you linked to, because it seems to be standard across repos for Codes of Conduct.


Pix1234:
AFAIK attribution isnt needed but let me talk with some people then i'll get back to this pr.

halfak:
OK so I think I'm with @Pix1234 here.  I don't see the value in adding such a large and complex attribution, but I think it would be great if we could add something like a summary of main points.  

Either way, thank you @melodykramer for your PR.  It's always wonderful to have contributions to review even if we don't end up accepting them.

melodykramer:
No prob! Good to get my GH fingers moving again! Until soon, friends!

PR #215, opened 2017-07-06
Switch ThreadingTimeout for SignalTimeout

halfak:
ThreadingTimeout can't block a long running regular expression, but SignalTimeout can.  Also this seems to work nicely within celery.

### DIFF
--- a/ores/scoring_systems/tests/test_celery_queue.py
+++ b/ores/scoring_systems/tests/test_celery_queue.py
@@ -1,5 +1,4 @@
 import celery
-import celerytest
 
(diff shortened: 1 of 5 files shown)


adamwight:
The code looks reasonable, but there are still two test failures (aside from flake8).

halfak:
I see!  It turns out that signal only works with whole numbers of seconds, so I switched to whole numbers in our tests and that fixed everything. :)

halfak:
Arg.  Just the flake8 errors.  Those should be cleaned up now.

PR #221, opened 2017-07-25
Adds stress_test utility

halfak:
Careful testing this.  Default values are benign.

### DIFF
--- a/ores/utilities/stress_test.py
+++ b/ores/utilities/stress_test.py
@@ -1,0 +1,168 @@
+"""
+Runs a stress test against an ORES api by sending it requests at a configurable
+rate.
+
+:Usage:
+    stress_test -h | --help
+    stress_test <context> <ores-url>...
+                [--model=<name>]
+                [--batch-size=<n>]
+                [--delay=<secs>]
+                [--input=<path>]
+                [--debug]
+                [--verbose]
+
+:Options:
+    -h --help          Prints this documentation
+    <ores-url>         URL of base ORES webserver.  If multiple URLS are
+                       provided, they will be cycled through round-robin.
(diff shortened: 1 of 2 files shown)


ores/utilities/stress_test.py
```diff
@@ -0,0 +1,168 @@
...
+    with ThreadPoolExecutor(max_workers=REQUEST_THREADS) as executor:
+        try:
+            while time.sleep(delay) is None:
+                ores_url = next(url_cycle)
+                rev_id_batch = list(islice(rev_ids_cycle, batch_size))
+                logger.debug("Submitting {0} rev_ids to {1}...".format(
+                    len(rev_ids), ores_url))
+                executor.submit(request_and_stat, ores_url, context,
+                                models, rev_id_batch, stats, verbose)
+        except KeyboardInterrupt:
```
adamwight: Is ^C the only way out?  Eventually we might want a --duration argument.

ores/utilities/stress_test.py
```diff
@@ -0,0 +1,168 @@
...
+        return
+
+    logger.debug("{0} response for {1} rev_ids...".format(
+        response.status_code, len(rev_ids)))
+    stats['total_responses'] += 1
+    stats['responses'][response.status_code] += 1
+
+    if response.status_code >= 200 and response.status_code < 300:
+        sys.stderr.write(".")
+        sys.stderr.flush()
```
adamwight: I would skip the flush, or only flush every 0=n mod 100.  Printing to console is incredibly expensive and will probably slow down the script.  Not sure how this will interact across threads, either...  Probably safer to do this with a https://docs.python.org/3/library/queue.html ?

PR #224, opened 2017-09-12
Use pytest instead of nosetests

Ladsgroup:
It's not all tests but it covers most

Bug: T174660


Pix1234:
With travis ci still failing it makes me uncomfortable to even endorse merging this, i would make sure prior to a merge to fix travis tests to:
A) Take advantage of the pytest switch
B) remove/fix/modify whats causing travis to fail

Ladsgroup:
Dude, Travis fails because of flake8 error (linting) and not an actual error.

Ladsgroup:
I was able to get coverage thing working too. \o/ Now there is no blocker issue. The code coverage is dropping strangely but that's something else that I can check later on.

PR #232, opened 2017-11-09
Revid stats

adamwight:


ores/scoring_systems/tests/util.py
```diff
@@ -49,6 +49,7 @@ def test_scoring_system(scoring_system):
         ScoreRequest("fakewiki", [1], ["fake"],
                      injection_caches={1: {wait_time: 0.05}},
                      model_info=['version']))
+    print(response.errors)
```
Ladsgroup: This doesn't look right, it should have some notes or something like that.

ores/scoring/models/rev_id_scorer.py
```diff
@@ -62,6 +64,17 @@ def score(self, feature_values):
             }
         }
 
+    def quick_statistics(self):
+        '''Jam some data through to generate statistics'''
```
Ladsgroup: The docstring is incorrect and violates PEP257, it should be "Jam some data through to generate statistics."
Also name of the method is better to be a verb.

ores/scoring_systems/tests/util.py
adamwight: Thanks, I commented.

ores/scoring/models/rev_id_scorer.py
adamwight: Good idea!

PR #233, opened 2017-11-13
Celery 4

adamwight:


Ladsgroup:
I'm doing this in incremental steps and in some parts in different approaches.

PR #237, opened 2017-11-30
T181693: make python 3 a hard requirement

Pix1234:


ores/__init__.py
```diff
@@ -1,5 +1,10 @@
+import sys
+import platform
 from .about import (__author__, __author_email__, __description__, __name__,
                     __url__, __version__)
 
 __all__ = [__name__, __version__, __author__, __author_email__,
            __description__, __url__]
+if sys.version_info <= (3, 0):
+    from pkg_resources import VersionConflict
+    raise VersionConflict("ORES requires Python '>=3' but your Python version is " + platform.python_version())
```
halfak: We'll need to do this check before any of the other imports.  This will make flake8 upset so I think we'll also need to make use of some `#noqa`'s

halfak:
Needs a line for "python_requires" in `setup()`

adamwight:
That worked perfectly, for both pip and setup.py!

I want to take a quick look at the #noqa to see if there's a workaround.

adamwight:
Thanks, looks great!

PR #243, opened 2017-12-20
Always configure logging; log to console

adamwight:
Push logging config down so it's executed from every entry point.
Detect whether we're running from the console, and mirror to console
if so.

Bug: https://phabricator.wikimedia.org/T182614

### DIFF
--- a/ores/applications/util.py
+++ b/ores/applications/util.py
@@ -3,8 +3,11 @@ import os.path
 import yamlconf
 import logging
 import logging.config
+import sys
 
 DEFAULT_DIRS = ["config/", "/etc/ores/"]
+DEFAULT_LOGGING_CONFIG = "logging_config.yaml"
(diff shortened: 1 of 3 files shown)


ores/applications/util.py
```diff
@@ -30,25 +33,40 @@ def build_config(config_dirs=DEFAULT_DIRS):
...
+            logging_config = DEFAULT_LOGGING_CONFIG
+
     if logging_config is not None:
         with open(logging_config) as f:
             logging_config = yamlconf.load(f)
             logging.config.dictConfig(logging_config)
 
+        # Secret sauce: if running from the console, mirror logs there.
+        if sys.stdin.isatty():
+            handler = logging.StreamHandler(stream=sys.stdout)
```
halfak: sys.stderr might be better

adamwight: Good idea—I also confirmed that this is the default behavior of logging.basicConfig, see https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L956

PR #245, opened 2018-01-04
Extends revscoring requirement to include 2.1.x

halfak:


.travis.yml
```diff
@@ -20,6 +20,7 @@ before_install:
 install:
   - pip install pip --upgrade
   - pip install wheel
+  - pip install nltk
```
adamwight: Why?  nltk is already included in requirements.txt

ores/metrics_collectors/util.py
```diff
@@ -1,6 +1,4 @@
 
 
 def format_set(s):
-    l = list(s)
-    l.sort()
-    return "{" + ", ".join(repr(value) for value in l) + "}"
+    return "{" + ", ".join(repr(value) for value in sorted(s)) + "}"
```
adamwight: ~~Isn't this just, json.dumps(sorted(s))?  Or does the formatting have to be more specific than that?~~

Oh, weird brace language.  NVM.

ores/utilities/stress_test.py
```diff
@@ -127,7 +127,7 @@ def request_and_stat(ores_url, context, models, rev_ids, stats, verbose):
     try:
         response = requests.get(
             ores_url + path, params=params)
-    except:
+    except Exception:
```
adamwight: That is.... bending the rules.  This would be a good place to catch the more specific exception, IMO.

requirements.txt
```diff
@@ -10,4 +10,4 @@ stopit >= 1.1.1, < 1.1.999
 yamlconf >= 0.2.2, < 0.2.999
 pytest >= 3.2.2, < 3.2.999
 pyyaml == 3.11
-revscoring >= 2.0.2, < 2.0.999
+revscoring >= 2.0.2, < 2.1.999
```
adamwight: You don't want a minimum version of 2.1.x?

.travis.yml
halfak: Because we call it directly below.  If it's required at this level, it should be installed at this level.

requirements.txt
halfak: Minimum version of 2.1.x?  what do you mean?

ores/utilities/stress_test.py
halfak: This is identical behavior while addressing the flake8 error.  Seems like improving the error handling in stress_test.py should be left for future work.  In this case, I don't think it matters much given that it's just a test.

ores/metrics_collectors/util.py
halfak: Right.  Should be deterministic.  Formatting JSON is not deterministic.

requirements.txt
Ladsgroup: I think awight means we drop support of 2.0.x

halfak: Ahh.  I don't see that as necessary in order to build models that are compatible, but I'm OK with it if y'all feel strongly.

Ladsgroup: I don't feel anything right now, The coffee hasn't kicked in yet.

PR #258, opened 2018-08-17
Add poolcounter class as the lock manager

Ladsgroup:
Bug: T201823

### DIFF
--- a/ores/lock_manager/lock_manager.py
+++ b/ores/lock_manager/lock_manager.py
@@ -3,13 +3,33 @@
 
 class LockManager:
     def connect(self):
+        """
+        Connect to the lock manager
+
+        returns false if it wasn't able to connect, true otherwise.
+        """
         raise NotImplementedError
 
-    def acq4me(self, key, workers, maxqueue, timeout):
+    def lock(self, key, workers, maxqueue, timeout):
+        """
+        Acquire a lock in the lock manager
+
+        returns false if it wasn't able to lock, true otherwise.
+        """
         raise NotImplementedError
 
     def release(self, key):
+        """
+        Release the lock in the lock manager
+
+        returns false if it wasn't able to release, true otherwise.
+        """
         raise NotImplementedError
 
     def close(self):
+        """
+        Close the connection to the lock manager
+
+        returns false if it wasn't able to close the connection, true otherwise.
+        """
         raise NotImplementedError
--- a/ores/lock_manager/lock_manager.py
+++ b/ores/lock_manager/lock_manager.py
@@ -1,0 +1,15 @@
+"""Lock manager interface."""
+
+
+class LockManager:
+    def connect(self):
+        raise NotImplementedError
+
+    def acq4me(self, key, workers, maxqueue, timeout):
+        raise NotImplementedError
+
+    def release(self, key):
+        raise NotImplementedError
+
+    def close(self):
+        raise NotImplementedError
--- a/ores/lock_manager/lock_manager.py
+++ b/ores/lock_manager/lock_manager.py
@@ -4,7 +4,7 @@
 class LockManager:
     def connect(self):
         """
-        Connect to the lock manager
+        Connect to the lock manager.
 
         returns false if it wasn't able to connect, true otherwise.
         """
@@ -12,7 +12,7 @@ class LockManager:
 
     def lock(self, key, workers, maxqueue, timeout):
         """
-        Acquire a lock in the lock manager
+        Acquire a lock in the lock manager.
 
         returns false if it wasn't able to lock, true otherwise.
         """
@@ -20,7 +20,7 @@ class LockManager:
 
     def release(self, key):
         """
-        Release the lock in the lock manager
+        Release the lock in the lock manager.
 
         returns false if it wasn't able to release, true otherwise.
         """
@@ -28,7 +28,7 @@ class LockManager:
 
     def close(self):
         """
-        Close the connection to the lock manager
+        Close the connection to the lock manager.
 
         returns false if it wasn't able to close the connection, true otherwise.
         """
--- a/ores/lock_manager/poolcounter.py
+++ b/ores/lock_manager/poolcounter.py
@@ -1,0 +1,58 @@
+"""
+Class to implement PoolCounter lock manager.
+
+This is useful to make sure too many connections are not coming from one IP.
+"""
+
+import socket
+
+from .lock_manager import LockManager
+
+
+class PoolCounter(LockManager):
+    def __init__(self, config, logger):
+        self.server = config['server']
+        self.port = config['port']
+        self.stream = None
+        self.logger = logger
+
+    def connect(self):
+        self.stream = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+        return self.stream.connect((self.server, self.port))
+
+    def acq4me(self, key, workers, maxqueue, timeout):
+        if not self.stream:
+            self.connect()
+        try:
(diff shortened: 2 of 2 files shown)


ores/lock_manager/poolcounter.py
```diff
@@ -0,0 +1,58 @@
...
+            data = self.stream.recv(4096).decode('utf-8')
+            print(data)
+        except socket.error as e:
+            self.stream = None
+            raise e
+
+        return data == 'LOCKED\n'
+
+    def release(self, key):
+        if not self.stream:
```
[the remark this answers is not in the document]
Ladsgroup: Yes, but this sanity check makes sure we don't try to run "send" method on None object which causes the whole thing to fall apart. It's better to be there IMO (Also it's copied from Gilles' version on Django: https://github.com/wikimedia/operations-debs-python-thumbor-wikimedia/blob/master/wikimedia_thumbor/poolcounter/__init__.py )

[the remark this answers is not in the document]
adamwight: I haven't interacted with PoolCounter personally, but the wikitech [page](https://wikitech.wikimedia.org/wiki/PoolCounter) says,

> The network protocol is line-based, with parameters separated by spaces (spaces in parameters are percent-encoded). The client opens a connection, sends a lock acquire command, does the work, sends a lock release command, then closes the connection. 

Maybe the docs are wrong?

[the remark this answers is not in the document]
adamwight: Sorry, I see what you're saying now.  If the connection was dropped, then the docs are ambiguous about whether we need to reconnect and release, or if the server will release automatically.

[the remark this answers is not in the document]
Ladsgroup: This page is trying to be the client so the responsibility of opening the connection, sending things, and closing it is on this class.

ores/lock_manager/poolcounter.py
```diff
@@ -0,0 +1,58 @@
...
+            self.connect()
+
+        try:
+            self.stream.send(bytes('RELEASE %s\n' % key, 'utf-8'))
+            data = self.stream.recv(4096).decode('utf-8')
+        except socket.error as e:
+            self.stream = None
+            raise e
+
+        return data == 'RELEASED\n'
```
adamwight: Seems safer to compare data.strip() == 'RELEASED'

ores/lock_manager/poolcounter.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def close(self):
+        if self.stream:
+            self.stream.close()
+            self.stream = None
+            return True
+        else:
+            return False
+
+    def debug(self, message):
```
adamwight: Shouldn't be merged

ores/lock_manager/lock_manager.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class LockManager:
+    def connect(self):
+        raise NotImplementedError
+
+    def acq4me(self, key, workers, maxqueue, timeout):
+        raise NotImplementedError
+
+    def release(self, key):
```
adamwight: The interface should document the return value

ores/lock_manager/poolcounter.py
Ladsgroup: The whole debug should be dropped :D

ores/lock_manager/poolcounter.py
```diff
@@ -0,0 +1,58 @@
...
+        self.stream = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+        return self.stream.connect((self.server, self.port))
+
+    def acq4me(self, key, workers, maxqueue, timeout):
+        if not self.stream:
+            self.connect()
+        try:
+            self.stream.send(bytes('ACQ4ME %s %d %d %d\n' % (key, workers, maxqueue, timeout), 'utf-8'))
+            data = self.stream.recv(4096).decode('utf-8')
+            print(data)
```
adamwight: if debug

ores/lock_manager/poolcounter.py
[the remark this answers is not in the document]
tgr: The server uses open connections as lock identifiers. I don't think it has any way to correlate two connections coming from the same client. So reconnecting and releasing (disconnecting) is just a costly noop, you could just return true instead.

Although arguably you should throw an exception since it signifies some kind of logic error in the code (trying to release a lock that was never obtained).

ores/lock_manager/lock_manager.py
```diff
@@ -0,0 +1,15 @@
+"""Lock manager interface."""
+
+
+class LockManager:
+    def connect(self):
+        raise NotImplementedError
+
+    def acq4me(self, key, workers, maxqueue, timeout):
```
adamwight: Rename "lock" to keep the interface generic.

Shouldn't all arguments after "key" come from config?  Not sure how I feel about "key", but it can stay here for now.

ores/lock_manager/poolcounter.py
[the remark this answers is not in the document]
adamwight: > Although arguably you should throw an exception since it signifies some kind of logic error in the code (trying to release a lock that was never obtained).

I like this idea, but @Ladsgroup is trying to *not* blow up because higher-level code should be the judge of whether its response will be valid, I think.

ores/lock_manager/poolcounter.py
```diff
@@ -0,0 +1,54 @@
...
+        except socket.error as e:
+            self.stream = None
+            raise e
+
+        return data.strip() == 'LOCKED'
+
+    def release(self, key):
+        if not self.stream:
+            return False
```
adamwight: Great, this was an elegant compromise.

ores/lock_manager/poolcounter.py
```diff
@@ -0,0 +1,54 @@
...
+
+        try:
+            self.stream.send(bytes('RELEASE %s\n' % key, 'utf-8'))
```
adamwight: Normally this would be `format` in py 3, but now I've typed more characters than you saved ;-)

ores/lock_manager/poolcounter.py
```diff
@@ -0,0 +1,53 @@
...
+        return self.stream.connect((self.server, self.port))
+
+    def lock(self, key, workers, maxqueue, timeout):
+        if not self.stream:
+            self.connect()
+        try:
+            self.stream.send(bytes('ACQ4ME %s %d %d %d\n' % (key, workers, maxqueue, timeout), 'utf-8'))
+            data = self.stream.recv(4096).decode('utf-8')
+        except socket.error as e:
+            self.stream = None
```
adamwight: Maybe we need to disconnect to avoid using up sockets?

adamwight: I think the best solution would be to use socket.socket as a context manager rather than storing as a class member.

Ladsgroup: Done

ores/lock_manager/poolcounter.py
```diff
@@ -26,7 +26,7 @@ def lock(self, key, workers, maxqueue, timeout):
             self.stream.send(bytes('ACQ4ME %s %d %d %d\n' % (key, workers, maxqueue, timeout), 'utf-8'))
             data = self.stream.recv(4096).decode('utf-8')
         except socket.error as e:
-            self.stream = None
+            self.close()
```
adamwight: Needs implementation!

adamwight: whoa, github user fail.  This is done.

PR #260, opened 2018-08-23
Use PoolCounter as lock manager of request IPs

Ladsgroup:
Todo:
 - Pass IP
 - Pass config for number of workers, etc.

Bug: T201823

### DIFF
--- a/config/00-main.yaml
+++ b/config/00-main.yaml
@@ -26,9 +26,7 @@ score_caches:
 #Lock managers
 lock_managers:
   pool_counter:
-    primary_node:
-      server: localhost
-      port: 7531
+    - localhost:7531
 
 # Metrics collection options
 metrics_collectors:
@@ -40,7 +38,6 @@ scoring_systems:
   defaults:
     metrics_collector: local_logging
     score_cache: memory_lru
-    lock_manager: pool_counter
     scoring_contexts:
       - testwiki
     timeout: 15 # seconds
--- a/config/00-main.yaml
+++ b/config/00-main.yaml
@@ -23,6 +23,12 @@ score_caches:
     prefix: ores
     socket_timeout: 15 # seconds
 
+#Lock managers
+lock_managers:
+  pool_counter:
+    server: localhost
+    port: 7531
+
 # Metrics collection options
 metrics_collectors:
   local_logging:
@@ -33,6 +39,7 @@ scoring_systems:
   defaults:
     metrics_collector: local_logging
     score_cache: memory_lru
+    lock_manager: pool_counter
     scoring_contexts:
       - testwiki
     timeout: 15 # seconds
--- a/config/00-main.yaml
+++ b/config/00-main.yaml
@@ -26,8 +26,9 @@ score_caches:
 #Lock managers
 lock_managers:
   pool_counter:
-    server: localhost
-    port: 7531
+    primary_node:
+      server: localhost
+      port: 7531
 
 # Metrics collection options
 metrics_collectors:
@@ -43,6 +44,8 @@ scoring_systems:
     scoring_contexts:
       - testwiki
     timeout: 15 # seconds
+    connections_per_ip: 4
+    connections_per_ip_hard: 7
   local_single_thread:
     class: ores.scoring_systems.SingleThread
   local_process_pool:
--- a/ores/lock_manager/poolcounter.py
+++ b/ores/lock_manager/poolcounter.py
@@ -4,26 +4,41 @@ Class to implement PoolCounter lock manager.
 This is useful to make sure too many connections are not coming from one IP.
 """
 
+import logging
 import socket
 
 from .lock_manager import LockManager
 from ..errors import TimeoutError
 
+logger = logging.getLogger(__name__)
+
+
 class PoolCounter(LockManager):
-    def __init__(self, config):
-        self.server = config['server']
-        self.port = config['port']
+    def __init__(self, primary_node, secondary_node=None):
+        self.primary_node = primary_node
+        self.secondary_node = secondary_node
         self.stream = None
 
     def connect(self):
         self.stream = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
-        return self.stream.connect((self.server, self.port))
+        try:
+            self.stream.connect(self.primary_node)
+        except ConnectionRefusedError:
+            if self.secondary_node:
+                logger.warning(
+                    'Can not connect to the primary PoolCounter node, '
+                    'falling back to the secondary node')
+                self.stream.connect(self.secondary_node)
+            else:
+                raise
 
     def lock(self, key, workers, maxqueue, timeout):
         if not self.stream:
             self.connect()
         try:
-            self.stream.send(bytes('ACQ4ME %s %d %d %d\n' % (key, workers, maxqueue, timeout), 'utf-8'))
+            self.stream.send(bytes(
+                'ACQ4ME %s %d %d %d\n' % (key, workers, maxqueue, timeout),
+                'utf-8'))
             data = self.stream.recv(4096).decode('utf-8')
         except socket.error as e:
             self.close()
@@ -52,9 +67,19 @@ class PoolCounter(LockManager):
             self.stream.close()
             self.stream = None
             return True
-        else:
-            return False
+
+        return False
 
     @classmethod
     def from_config(cls, config, name, section_key="lock_managers"):
-        return cls(config[section_key][name])
+        primary_node_config = config[section_key][name]['primary_node']
+        secondary_node_config = config[section_key][name].get('secondary_node')
+        if secondary_node_config:
+            secondary_node = (secondary_node_config['server'],
+                              secondary_node_config['port'])
+        else:
+            secondary_node = None
+        return cls(
+            (primary_node_config['server'], primary_node_config['port']),
+            secondary_node
+        )
--- a/ores/lock_manager/poolcounter.py
+++ b/ores/lock_manager/poolcounter.py
@@ -7,7 +7,7 @@ This is useful to make sure too many connections are not coming from one IP.
 import socket
 
 from .lock_manager import LockManager
-
+from ..errors import TimeoutError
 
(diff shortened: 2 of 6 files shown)


Ladsgroup:
Now it works fine, I tested locally. It doesn't pass IP yet but I rather do it in another PR as it touches lots of files.

ores/lock_manager/poolcounter.py
```diff
@@ -4,31 +4,49 @@
...
+from ..errors import TimeoutError
+
+logger = logging.getLogger(__name__)
 
 
 class PoolCounter(LockManager):
-    def __init__(self, config):
-        self.server = config['server']
-        self.port = config['port']
+    def __init__(self, primary_node, secondary_node=None):
```
adamwight: We discussed in IRC, but for the record https://www.mediawiki.org/wiki/Extension:PoolCounter#Configuration suggests a different implementation, where we have a cluster of pool counter servers and choose a server based on consistent hashing.  The approach here should still work, however.

ores/lock_manager/poolcounter.py
```diff
@@ -49,5 +67,19 @@ def close(self):
...
+        primary_node_config = config[section_key][name]['primary_node']
+        secondary_node_config = config[section_key][name].get('secondary_node')
+        if secondary_node_config:
+            secondary_node = (secondary_node_config['server'],
+                              secondary_node_config['port'])
         else:
-            return False
+            secondary_node = None
+        return cls(
+            (primary_node_config['server'], primary_node_config['port']),
```
adamwight: It's anxiety-inducing that the primary and secondary clauses aren't parallel.

config/00-main.yaml
```diff
@@ -23,6 +23,13 @@ score_caches:
     prefix: ores
     socket_timeout: 15 # seconds
 
+#Lock managers
```
adamwight: Doesn't seem like this should be the default, since most small installations will probably have a single client, and might intentionally want to allow max thoroughput.

ores/scoring_systems/scoring_system.py
```diff
@@ -17,12 +18,16 @@
 class ScoringSystem(dict):
 
     def __init__(self, context_map, score_cache=None,
-                       metrics_collector=None, timeout=None):
+                       metrics_collector=None, timeout=None, lock_manager=None,
+                       connections_per_ip=4, connections_per_ip_hard=7):
```
adamwight: Feels like we these last parameters belong to the lock_manager and shouldn't be exposed to ScoringSystem.

ores/scoring_systems/scoring_system.py
```diff
@@ -33,8 +38,15 @@ def check_context_models(self, request):
         if len(missing_models) > 0:
             raise MissingModels(request.context_name, missing_models)
 
-    def score(self, request):
+    def score(self, request, ip=None):
```
adamwight: IP should be included in ScoreRequest.

ores/scoring_systems/scoring_system.py
```diff
@@ -240,6 +255,35 @@ def _lookup_cached_score(self, request, rev_id, model_name):
...
+    def _connect_to_lock_manager(self):
+        try:
+            self.lock_manager.connect()
+        # Lock manager is down, let's do nothing
+        except ConnectionRefusedError:
+            logger.warning('Can not connect to lock manager')
+            return False
+        return True
+
+    def _lock_ip(self, ip):
```
adamwight: IMO this should be generalized, so we can lock on things other than just IP.  For example, we might have a separate pool with different configuration for requests that include features.

ores/lock_manager/poolcounter.py
akosiaris: Yeah both fallback and consistent hashing would work. Consistent hashing is preferable as far as I am concerned, but I am fine with fallback for now.

config/00-main.yaml
Ladsgroup: Yes, I think we should remove it from defaults in below

ores/lock_manager/poolcounter.py
Ladsgroup: okay, I implement consistent hashing here.

ores/scoring_systems/scoring_system.py
Ladsgroup: no, lock manager can be used for anything, for example lock manager can be used to lock a certain score so other workers don't score it. In general it's a little bit YAGNI, but it shows these are conceptually different. You can see that Thumbor also followed this line of thought which I agree with.

ores/scoring_systems/scoring_system.py
Ladsgroup: Yes, let me fix that.

ores/lock_manager/poolcounter.py
Ladsgroup: Done now

Ladsgroup:
All stuff has been addressed \o/ Tested locally and it works.

ores/scoring_systems/scoring_system.py
```diff
@@ -38,14 +38,13 @@ def check_context_models(self, request):
...
-    def score(self, request, ip=None):
+    def score(self, request):
         self.check_context_models(request)
 
         locked = None
-        if ip and (not request.precache) and self.lock_manager is not None:
-            connected = self._connect_to_lock_manager()
-            if connected:
-                locked = self._lock_ip(ip.strip())
+        if request.ip and (not request.precache) and \
```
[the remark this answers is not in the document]
Ladsgroup: Yeah, we discussed this halfak and decided to use precache flag instead of the whitelisted IPs because all of DoSes we have had until now were not intentional (they weren't malicious)

ores/lock_manager/poolcounter.py
```diff
@@ -14,27 +15,31 @@
...
+        self.nodes = nodes
         self.stream = None
 
-    def connect(self):
-        self.stream = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+    def connect(self, key):
+        hashes = []
+        for node in self.nodes:
+            hashes.append(
+                (node, hashlib.sha256(bytes(node[0] + key, 'utf-8')).hexdigest())
```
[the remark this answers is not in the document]
Ladsgroup: It's based on the hashing system the extension use

ores/lock_manager/poolcounter.py
```diff
@@ -14,27 +15,31 @@
...
+    def __init__(self, nodes):
+        self.nodes = nodes
         self.stream = None
 
-    def connect(self):
-        self.stream = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+    def connect(self, key):
+        hashes = []
+        for node in self.nodes:
+            hashes.append(
```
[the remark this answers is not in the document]
Ladsgroup: Alex said above that he prefers the distribution and not the fallback system.

ores/lock_manager/poolcounter.py
```diff
@@ -14,27 +15,31 @@
...
+                (node, hashlib.sha256(bytes(node[0] + key, 'utf-8')).hexdigest())
+            )
+        node = sorted(hashes, key=lambda i: i[1])[0][0]
         try:
-            self.stream.connect(self.primary_node)
+            self.stream = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
```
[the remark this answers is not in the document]
Ladsgroup: Yeah but lock managers are not that important specially in this case, losing them means nothing here.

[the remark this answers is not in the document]
adamwight: What worries me is that an abandoned socket will probably be left open, and the total socket allowance is usually pretty low.  It looks like Python will garbage collect, but that's unpredictable and may keep sockets open on the other end: https://docs.python.org/2/howto/sockets.html#disconnecting

ores/lock_manager/poolcounter.py
```diff
@@ -4,31 +4,54 @@
...
+                (node, hashlib.sha256(bytes(node[0] + key, 'utf-8')).hexdigest())
+            )
+        node = sorted(hashes, key=lambda i: i[1])[0][0]
+        try:
+            self.stream = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+            self.stream.connect(node)
+        except ConnectionRefusedError:
+            logger.warning(
+                'Can not connect to the PoolCounter node %s' % node[0])
+            return False
```
adamwight: We could also fall back to the next server, but I'm sure this is fine for now.

ores/lock_manager/poolcounter.py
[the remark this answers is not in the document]
Ladsgroup: My knowledge about TCP is pretty novice but AFAIK poolcounter has timeout and closes connections if they left open more than a certain time.

ores/lock_manager/poolcounter.py
Ladsgroup: we return false and upstream service just ignores that.

ores/lock_manager/poolcounter.py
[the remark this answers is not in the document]
adamwight: Gotcha, I found something different in mw-core but now I see PoolCounter_ConnectionManager.php.

ores/lock_manager/poolcounter.py
adamwight: Alright!

legoktm:
I didn't see any support for ACQ4ANY...any reason for that? It seems like most of the requests could share caches, right?

Ladsgroup:
@legoktm I might be understanding the documentation incorrectly but this poolcounter implementation here is basically just a lock manager based on IP adrress so if we get seven requests at the same from on IP, we should start emitting TImeOutError and after four requests, it should respond slower so I went with ACQ4ME because "When a lock of this kind is released, only one waiting process will be woken, so as to keep the worker population the same." meaning if there is six connection and two are waiting, if one process is done and asked to release the lock, we only allow one more process to continue and not two. Did I misunderstand the PoolCounter here?

PR #261, opened 2018-09-05
Send timing of locks to metrics collector

Ladsgroup:


ores/scoring_systems/scoring_system.py
```diff
@@ -44,7 +44,12 @@ def score(self, request):
         locked = None
         if request.ip and (not request.precache) and \
                 self.lock_manager is not None:
+            start = time.time()
             locked = self._lock_ip(request.ip)
+            self.metrics_collector.send_timing_event(
+                'poolcounter',
+                'locking_response_time',
+                duration=time.timte() - start)
```
halfak: This should use the standard interface for a metrics_collector: https://github.com/wikimedia/ores/blob/7cfe63eee306975b9432c89538aa360a28e9536c/ores/metrics_collectors/metrics_collector.py

Ladsgroup: Done

halfak: `timte` is a typo

PR #262, opened 2018-09-19
Move releasing the lock to after doing the actual job

Ladsgroup:
Right now, it basically connects to PoolCounter, lock it,
and then immediately releases it basically making it completely useless.

This bug was introduced by yours truly


adamwight:
Here's to many more ;-)

PR #264, opened 2018-09-20
Return 504 instead of 408 when worker times out

Ladsgroup:
4xx codes means there is something wrong with the client
but this it's something wrong with the server.

Change-Id: I6d9f981da4c56f4737492c016a96ad8be0b96088


adamwight:
Fancy and looks correct!

PR #266, opened 2018-09-21
Send stats to metrics collectors when the response is not 200

Ladsgroup:
Bug: T205051


halfak:
Hmm.  I don't like that we abstract away the response code in the response formatters but it is hard-coded in the logging.  I wonder if we could create some common variables for the two.  E.g. `responses.codes.SERVER_OVERLOADED = 503`

Ladsgroup:
Done, Will add more constants for other responses later. For now, only the ones we are sending the metrics.

PR #268, opened 2018-09-28
Add LogstashHandler and LogstashFormatter

Ladsgroup:
Mostly cleaned up version of python-logstash library

Bug: T181546
Bug: T181630
Bug: T169586
Bug: T205256
Bug: T168921

### DIFF
--- a/ores/logging/logstash_fomatter.py
+++ b/ores/logging/logstash_fomatter.py
@@ -1,0 +1,87 @@
+import traceback
+import logging
+import socket
+from datetime import datetime
+import json
+
+
+class LogstashFormatter(logging.Formatter):
+
+    def __init__(self, message_type='Logstash', tags=None, host=None):
+        self.message_type = message_type
+        self.tags = tags if tags is not None else []
+        self.host = host if host is not None else socket.gethostname()
+
+    def format(self, record):
+        # Create message dict
+        message = {
+            '@timestamp': self.format_timestamp(record.created),
+            '@version': '1',
+            'message': record.getMessage(),
+            'host': self.host,
+            'path': record.pathname,
+            'tags': self.tags,
+            'type': self.message_type,
+
+            # Extra Fields
+            'level': record.levelname,
+            'logger_name': record.name,
+        }
+
+        # Add extra fields
+        message.update(self.get_extra_fields(record))
+
+        # If exception, add debug info
+        if record.exc_info:
+            message.update(self.get_debug_fields(record))
+
+        return self.serialize(message)
+
+    def get_extra_fields(self, record):
+        # The list contains all the attributes listed in
+        # http://docs.python.org/library/logging.html#logrecord-attributes
+        skip_list = (
+            'args', 'asctime', 'created', 'exc_info', 'filename', 'funcName',
+            'levelname', 'levelno', 'lineno', 'message', 'module', 'msecs',
+            'msg', 'name', 'pathname', 'process', 'processName',
+            'relativeCreated', 'stack_info', 'thread', 'threadName')
+
+        easy_types = (str, bool, dict, float, int, list, type(None))
+
+        fields = {}
+
+        for key, value in record.__dict__.items():
+            if key not in skip_list:
+                if isinstance(value, easy_types):
+                    fields[key] = value
+                else:
+                    fields[key] = repr(value)
+
+        return fields
+
+    def get_debug_fields(self, record):
+        return {
+            'stack_trace': self.format_exception(record.exc_info),
+            'lineno': record.lineno,
+            'process': record.process,
+            'thread_name': record.threadName,
+            'funcName': record.funcName,
+            'processName': record.processName,
+        }
+
+    @classmethod
+    def format_source(cls, message_type, host, path):
+        return "%s://%s/%s" % (message_type, host, path)
+
+    @classmethod
+    def format_timestamp(cls, time):
+        tstamp = datetime.utcfromtimestamp(time)
+        return tstamp.strftime("%Y-%m-%dT%H:%M:%S") + ".%03d" % (tstamp.microsecond / 1000) + "Z"
+
+    @classmethod
+    def format_exception(cls, exc_info):
+        return ''.join(traceback.format_exception(*exc_info)) if exc_info else ''
+
+    @classmethod
+    def serialize(cls, message):
+        return bytes(json.dumps(message), 'utf-8')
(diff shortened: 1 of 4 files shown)


adamwight:
This would fit better as an independent library, forking python-logstash.

ores/logging/logstash_fomatter.py
```diff
@@ -0,0 +1,87 @@
...
+        # If exception, add debug info
+        if record.exc_info:
+            message.update(self.get_debug_fields(record))
+
+        return self.serialize(message)
+
+    def get_extra_fields(self, record):
+        # The list contains all the attributes listed in
+        # http://docs.python.org/library/logging.html#logrecord-attributes
+        skip_list = (
```
halfak: Should be a class variable.

ores/logging/logstash_fomatter.py
```diff
@@ -0,0 +1,87 @@
...
+            'args', 'asctime', 'created', 'exc_info', 'filename', 'funcName',
+            'levelname', 'levelno', 'lineno', 'message', 'module', 'msecs',
+            'msg', 'name', 'pathname', 'process', 'processName',
+            'relativeCreated', 'stack_info', 'thread', 'threadName')
+
+        easy_types = (str, bool, dict, float, int, list, type(None))
```
halfak: Should be a class variable

Ladsgroup:
> This would fit better as an independent library, forking python-logstash.

The reason I hasn't forked it is this: https://phabricator.wikimedia.org/T168921#4619397

ores/logging/logstash_fomatter.py
Ladsgroup: Done

ores/logging/logstash_fomatter.py
Ladsgroup: Done

ores/logging/logstash_fomatter.py
```diff
@@ -0,0 +1,85 @@
...
+    def format(self, record):
+        # Create message dict
+        message = {
+            '@timestamp': self.format_timestamp(record.created),
+            '@version': '1',
+            'message': record.getMessage(),
+            'host': self.host,
+            'path': record.pathname,
+            'tags': self.tags,
+            'type': self.message_type,
```
akosiaris: I am a bit confused about message_type. Will it always be "Logstash" ? and still sent over the wire ?

ores/logging/logstash_handler.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    Python logging handler for Logstash.
+
+    :param host: The host of the logstash server.
+    :param port: The port of the logstash server (default 5959).
+    :param message_type: The type of the message (default logstash).
+    :param tags: list of tags for a logger (default is None).
+    """
+
+    def __init__(self, host, port=5959, message_type='logstash', tags=None):
```
akosiaris: I don't think there is a default port for the udp protocol. GELF on the other hand which is what we use defaults to 12201 (https://www.elastic.co/guide/en/logstash/current/plugins-inputs-gelf.html) and we actually have it set to 12201.

ores/logging/logstash_fomatter.py
akosiaris: Btw, https://wikitech.wikimedia.org/wiki/Logstash#Gotchas

```
Make sure logging events sent to the GELF input don't have a "type" or "_type" field set, or if set, that it contains the value "gelf". The gelf/logstash config discards any events that have a different value set for "type" or "_type". The final "type" seen in Kibana/Elasticsearch will be take from the "facility" element of the original GELF packet. The application sending the log data to Logstash should set "facility" to a reasonably unique value that identifies your application.
```

Ladsgroup: Completely removed it.

ores/logging/logstash_handler.py
Ladsgroup: Changed the default, thanks!

ores/logging/logstash_handler.py
```diff
@@ -0,0 +1,21 @@
...
+        super(LogstashHandler, self).__init__(host, port)
+        self.formatter = LogstashFormatter(message_type, tags)
+
+    def makePickle(self, record):
```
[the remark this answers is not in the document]
adamwight: I think it would be nicer to override `emit()` since we're actually not pickling, see https://github.com/python/cpython/blob/master/Lib/logging/handlers.py#L620 for the base implementation.

ores/logging/tests/test_formatter.py
```diff
@@ -0,0 +1,29 @@
+from logging import LogRecord
+from ..logstash_fomatter import LogstashFormatter
+import json
+
+
+def test_format():
```
adamwight: Thank you for the test!

adamwight:
This looks like it's almost ready to merge.  Tests are failing with:
```
E       AssertionError: assert {'@timestamp'...t.wmnet', ...} == {'@timestamp':...t.wmnet', ...}
E         Omitting 9 identical items, use -vv to show
E         Right contains more items:
E         {'type': 'Logstash'}
```

ores/logging/logstash_fomatter.py
```diff
@@ -0,0 +1,83 @@
...
+        return {
+            'stack_trace': self.format_exception(record.exc_info),
+            'lineno': record.lineno,
+            'process': record.process,
+            'thread_name': record.threadName,
+            'funcName': record.funcName,
+            'processName': record.processName,
+        }
+
+    @classmethod
```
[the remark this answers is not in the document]
halfak: Good point.  Probably should be staticmethod or module-level.  Seems better to be staticmethod if they aren't going to be re-used.  And then they can be renamed to "_protected_dont_try_to_use_me_method()" style names.   From a performance perspective, it doesn't matter.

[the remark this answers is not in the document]
Ladsgroup: These are being reused by other codes that we are not using. i.e. LogHandlers need to implement these methods this way.

ores/logging/logstash_handler.py
```diff
@@ -0,0 +1,20 @@
+
```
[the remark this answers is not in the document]
Ladsgroup: Done

ores/logging/logstash_handler.py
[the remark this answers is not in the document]
Ladsgroup: It has the error handling.

ores/logging/logstash_fomatter.py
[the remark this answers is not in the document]
halfak: Aha!  It's strange that classmethod is what they chose because staticmethod should work.  But if that's how they set up the interface, it should be what we do.

PR #271, opened 2018-10-08
Make ScoreRequest JSON serialiazble

Ladsgroup:
Bug: T206334

### DIFF
--- a/ores/score_request.py
+++ b/ores/score_request.py
@@ -73,3 +73,27 @@ class ScoreRequest:
                 "injection_caches={0!r}".format(self.injection_caches),
                 "ip={0!r}".format(self.ip),
                 "model_info={0!r}".format(self.model_info)]))
+
+    def toJSON(self):
+        return {
+            'context': self.context_name,
+            'rev_ids': self.rev_ids,
+            'model_names': self.model_names,
+            'precache': self.precache,
+            'include_features': self.include_features,
+            'injection_caches': self.injection_caches,
+            'ip': self.ip,
+            'model_info': self.model_info
+        }
+
+    @classmethod
+    def fromJSON(cls, kwargs):
+        return cls(
+            kwargs['context'],
+            kwargs['rev_ids'],
+            kwargs['model_names'],
+            precache=kwargs.get('precache', False),
+            include_features=kwargs.get('include_features', False),
+            injection_caches=kwargs.get('injection_caches'),
+            model_info=kwargs.get('model_info'),
+            ip=kwargs.get('ip'))
--- a/ores/tests/test_score_request.py
+++ b/ores/tests/test_score_request.py
@@ -2,7 +2,7 @@ from ..score_request import ScoreRequest
 
 
 def test_score_request():
-    sr = ScoreRequest("foo", [1, 2, 3], ["bar", "baz"])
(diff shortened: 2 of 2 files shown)


Ladsgroup:
I tested on all ways possible and it worked just fine.

ores/score_request.py
```diff
@@ -73,3 +73,27 @@ def __repr__(self):
                 "injection_caches={0!r}".format(self.injection_caches),
                 "ip={0!r}".format(self.ip),
                 "model_info={0!r}".format(self.model_info)]))
+
+    def toJSON(self):
```
adamwight: I don't think "JSON" is the right word for what we're doing here.  Technically, it's "marshalling" although that doesn't exactly roll off the tongue.  Also, Python breaks the term by misusing it as another name for serialization.

What do you think about "toData" and "newFromData"?

ores/score_request.py
```diff
@@ -73,3 +73,27 @@ def __repr__(self):
...
+            'rev_ids': self.rev_ids,
+            'model_names': self.model_names,
+            'precache': self.precache,
+            'include_features': self.include_features,
+            'injection_caches': self.injection_caches,
+            'ip': self.ip,
+            'model_info': self.model_info
+        }
+
+    @classmethod
```
adamwight: Nice use of classmethod :-)

ores/score_request.py
```diff
@@ -73,3 +73,27 @@ def __repr__(self):
...
+    def fromJSON(cls, kwargs):
+        return cls(
+            kwargs['context'],
+            kwargs['rev_ids'],
+            kwargs['model_names'],
+            precache=kwargs.get('precache', False),
```
adamwight: Curious why there are defaults here.  It doesn't seem like the fields are optional in our serialized format, at least.

ores/score_request.py
halfak: Can you say more why you don't think "JSON" is the right term?  Maybe you mean "JSONable" since it's not turning the data structure into JSON, but rather a trivially JSON-able python data structure.

halfak: Oh also the name is a style issue.  Should probably be "to_json" or something like that.

ores/score_request.py
```diff
@@ -73,3 +73,27 @@ def __repr__(self):
...
+            'model_names': self.model_names,
+            'precache': self.precache,
+            'include_features': self.include_features,
+            'injection_caches': self.injection_caches,
+            'ip': self.ip,
+            'model_info': self.model_info
+        }
+
+    @classmethod
+    def fromJSON(cls, kwargs):
```
halfak: should probably be "from_json"

ores/score_request.py
halfak: Indeed.  Why not just do cls(**kwargs)? 

Also, I think this is a confusing use of "kwargs" since it generally refers to key word arguments, but here, you are taking a deserialized JSON document.

ores/score_request.py
Ladsgroup: Renamed them to to_json and from_json

ores/score_request.py
Ladsgroup: Got it from pywikibot 🙈🙈

ores/score_request.py
Ladsgroup: I renamed it to to_json

PR #273, opened 2018-10-12
Clean up timeout code

Ladsgroup:
- Remove unreachable code
 - Change print_tb (requires argument) to print_stack


halfak:
I see.  It looks like we can either have the context manager throw an exception or swallow it and check the context manager state later.

halfak:
Based on https://pypi.org/project/stopit/#timeout-context-manager

PR #280, opened 2018-11-02
Add connection timeout to PoolCounter

Ladsgroup:
Bug: T208577


halfak:
I want to let @akosiaris have a look.  But this LGTM.  Feel free to self-merge if alex doesn't get a chance.

ores/lock_manager/poolcounter.py
```diff
@@ -28,11 +29,17 @@ def connect(self, key):
         node = sorted(hashes, key=lambda i: i[1])[0][0]
         try:
             self.stream = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+            socket.settimeout(self.connection_timeout)
```
akosiaris: self.stream.settimeout(self.connection_timeout) I guess ?

ores/lock_manager/poolcounter.py
```diff
@@ -28,11 +29,17 @@ def connect(self, key):
...
             self.stream.connect(node)
+            socket.settimeout(None)
```
akosiaris: Why set back to blocking mode ?

Also shouldn't this be self.stream.settimeout(None) ?

ores/lock_manager/poolcounter.py
```diff
@@ -81,6 +88,7 @@ def close(self):
     @classmethod
     def from_config(cls, config, name, section_key="lock_managers"):
         nodes = []
+        # TODO: Fix config so we can inject connection_timeout
```
akosiaris: Yes please. It would be great for this to be configurable

ores/lock_manager/poolcounter.py
Ladsgroup: Yes (I'm so facepalm right now, how I made such a stupid mistake)

ores/lock_manager/poolcounter.py
Ladsgroup: https://stackoverflow.com/questions/3432102/python-socket-connection-timeout

ores/lock_manager/poolcounter.py
Ladsgroup: Yes, let's fix this issue at hand first. I make a phabricator ticket.

ores/lock_manager/poolcounter.py
akosiaris: Sure, but where exactly is the corresponding call to `socket.makefile()` that requires the socket to be in blocking mode ?

Ladsgroup: hmm, my original intention is to avoid making the socket to return timeout while connected when it has to wait for let's say one second because poolcounter is keeping the client waiting due to workers being busy (the soft threshold). My knowledge in this area is pretty novice, wouldn't cause issues if we remove the blocking part? if no, I definitely remove it.

akosiaris: It's a timeout per operation, not a global timeout. So `connect()` is an operation, `send()` is an operation and so on. And we can assume those 2 to succeed quickly in the WMF infra and under normal conditions. And if they any of the 2 operations timeout is actually good and a sane thing to close the connection and retry.

That being said, of course `recv()` is an operation as well and that is indeed going to timeout. And given the code below (L47) (`data = self.stream.recv(4096).decode('utf-8')`)is written to expect blocking behavior and not non-blocking behavior, it's indeed prudent to set again the blocking behavior, so fine by me.

That being said note that there exists an issue where the code might get stuck on trying to `send()` or `recv()` and fail when a poolcounter servers freezes (or is forcefully disconnected from the network). This will be picked up at the end by the kernel which will timeout the connection (after quite some time though). This will probably create a brief spike of multiple lock manager errors and locked processes. After that is done, on retry the offending poolcounter server will not be used. This will NOT happen when doing maintenance as the server process will be gracefully terminated and the connections will be terminated normally. 

As a general comment I would suggest never writing socket blocking code but rather use `select()` and read/write only when `select()` says it's ok to do so and then use nonblocking calls. Yes I know, socket programming is a pain.

akosiaris:
LGTM

ores/lock_manager/poolcounter.py
Ladsgroup: Thanks for all of this information. <3

PR #282, opened 2018-11-02
Add bc layer for ScoreRequest argument in score process task in celery

Ladsgroup:
After this get deployed, we will change the reqeust object to send it over json

Bug: T206333

### DIFF
--- a/ores/scoring_systems/celery_queue.py
+++ b/ores/scoring_systems/celery_queue.py
@@ -9,6 +9,7 @@ import celery.states
 import mwapi.errors
 import revscoring.errors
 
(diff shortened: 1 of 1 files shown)


ores/scoring_systems/celery_queue.py
```diff
@@ -58,6 +59,9 @@ def _process_score_map(request, model_names, rev_id, root_cache):
             logger.info("Generating a score map for {0}"
                         .format(request.format(rev_id, model_names)))
 
+            if not isinstance(request, ScoreRequest):
```
adamwight: Could comment that this is transition code and will be deprecated later.

PR #284, opened 2018-11-03
Log unknown errors

Ladsgroup:
Bug: T208623

### DIFF
--- a/ores/wsgi/responses.py
+++ b/ores/wsgi/responses.py
@@ -1,9 +1,12 @@
+import logging
+
 from flask_jsonpify import jsonify
 
 SERVER_OVERLOADED = 503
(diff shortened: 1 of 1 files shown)


ores/wsgi/responses.py
```diff
@@ -1,9 +1,12 @@
+import logging
+
 from flask_jsonpify import jsonify
 
 SERVER_OVERLOADED = 503
 NOT_FOUND = 404
 TIMEOUT = 504
 TOO_MANY_REQUESTS = 429
+logger = logging.getLogger('ores')
```
halfak: Should be `logger = logging.getLogger(__name__)`

halfak:
I added one note.

PR #290, opened 2018-11-16
Remove all of celery configs

Ladsgroup:
So we can fix them in puppet, deploy, then re-add them back here

Bug: T209587


Ladsgroup:
This would break almost all of third party installations that are depending on puppet. So I would suggest merging it and deploying at the time of deployment window and revert it back (with new configs) super quickly

Ladsgroup:
I will revert this asap

PR #300, opened 2018-12-13
Fix parsing of precache config

Ladsgroup:
We can have several models per event and context

Bug: T211267


halfak:
This is a nasty bug!  I think we'll want to have some tests for this.

PR #302, opened 2018-12-14
Make check for parent rev in precache more robust to change

Ladsgroup:
The stream can stop sending rev_parent_id for page creation or send null.
We should handle all possible cases

Bug: T211267


Pchelolo:
+1

PR #306, opened 2018-12-19
Add tests for redis task tracker and score cache

Ladsgroup:
Bug: T212090

### DIFF
--- a/ores/score_caches/tests/test_redis.py
+++ b/ores/score_caches/tests/test_redis.py
@@ -17,3 +17,30 @@ def cache():
 def test_lookup(cache):
     cache.store(0.421, 'testwiki', 'damaging', 124)
     assert cache.lookup('testwiki', 'damaging', 124) == 0.421
+
+
+class RedisMock():
+    def __init__(self):
+        self.values = {}
+
+    def setex(self, key, ttl, value):
+        self.values[key] = value
+        return True
+
+    def get(self, key):
+        return self.values.get(key)
+
+    def delete(self, key):
+        del self.values[key]
+        return True
+
+
+@pytest.fixture
+def cache_mock():
+    return Redis(RedisMock(), 0, 'foo')
+
+
+@pytest.mark.redis
+def test_lookup_mock(cache_mock):
+    cache_mock.store(0.421, 'testwiki', 'damaging', 124)
+    assert cache_mock.lookup('testwiki', 'damaging', 124) == 0.421
--- a/ores/score_caches/tests/test_redis.py
+++ b/ores/score_caches/tests/test_redis.py
@@ -1,11 +1,12 @@
 import pytest
 
-from ..redis import Redis
+from ..redis import Redis as RedisScoreCache
+from ...tests.redis_mock import RedisMock
 
 
 @pytest.fixture
 def cache():
-    return Redis.from_config({
+    return RedisScoreCache.from_config({
         'score_caches': {
             'redis': {
                 'host': 'localhost',
@@ -19,25 +20,9 @@ def test_lookup(cache):
     assert cache.lookup('testwiki', 'damaging', 124) == 0.421
 
 
-class RedisMock():
-    def __init__(self):
-        self.values = {}
-
-    def setex(self, key, ttl, value):
-        self.values[key] = value
-        return True
-
-    def get(self, key):
-        return self.values.get(key)
-
-    def delete(self, key):
-        del self.values[key]
-        return True
-
-
 @pytest.fixture
 def cache_mock():
-    return Redis(RedisMock(), 0, 'foo')
+    return RedisScoreCache(RedisMock(), 0, 'foo')
 
 
 @pytest.mark.redis
--- a/ores/score_caches/tests/test_redis.py
+++ b/ores/score_caches/tests/test_redis.py
@@ -13,6 +13,7 @@ def cache():
                 'ttl': 100}}}, 'redis')
 
 
+@pytest.mark.redis
 def test_lookup(cache):
-    cache.store(0.421, 'teswiki', 'damaging', 124)
-    assert cache.lookup('teswiki', 'damaging', 124) == 0.421
+    cache.store(0.421, 'testwiki', 'damaging', 124)
+    assert cache.lookup('testwiki', 'damaging', 124) == 0.421
--- a/ores/score_caches/tests/test_redis.py
+++ b/ores/score_caches/tests/test_redis.py
@@ -1,0 +1,18 @@
+import pytest
+
+from ..redis import Redis
+
+
+@pytest.fixture
+def cache():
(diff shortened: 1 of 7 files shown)


Ladsgroup:
Ignore the branch name, at first I wanted to write celery tests then I change my mind :D

ores/score_caches/tests/test_redis.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@pytest.fixture
+def cache():
+    return Redis.from_config({
+        'score_caches': {
+            'redis': {
+                'host': 'localhost',
+                'prefix': 'ores-derp',
+                'ttl': 100}}}, 'redis')
```
halfak: Assumes that redis is installed on the test machine.  Maybe we should catch the error we expect if redis is not available and raise a meaningful error message saying "Redis must be installed and available via 'localhost'".

Ladsgroup: The problem is that these fixtures only provide the class and don't do anything to check it against and see if redis is not setup. If you stop redis, tests fail with "redis.exceptions.ConnectionError: Error 111 connecting to localhost:6379. Connection refused." which is not that bad. One way to make it better to wrap all of redis tests into a try but I'm not sure if it's worth the trouble. I add pytest.mark('redis') to it so we can easily skip all those though. That would help a lot.

ores/score_caches/tests/test_redis.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@pytest.mark.redis
+def test_lookup(cache):
+    cache.store(0.421, 'teswiki', 'damaging', 124)
+    assert cache.lookup('teswiki', 'damaging', 124) == 0.421
```
halfak: "testwiki"

ores/task_tracker/redis_task_tracker.py
```diff
@@ -46,8 +46,8 @@ def from_parameters(cls, *args, ttl=None, prefix=None, **kwargs):
     @classmethod
     def from_config(cls, config, name, section_key="task_trackers"):
         """
-        score_caches:
-            task_trackers:
+        task_trackers:
+            redis:
                 class: ores.lock_manager.Redis
```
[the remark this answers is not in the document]
Ladsgroup: Done

ores/score_caches/tests/test_redis.py
Ladsgroup: done

ores/score_caches/tests/test_redis.py
halfak: This is complicated.  We don't usually rely on external services in unit tests because it complicates things.  What exactly are we trying to test here?  Could we test it without using the redis service directly?  We can always mock a redis connection pretty easily and check that connecting to redis actually works when we stand a real instance of the service up.

Ladsgroup: It will be super hard to implement such measures + all of these tests are marked as "redis" type so in case redis is not available you can just skip them by running "pytest -m "not redis"". We depend on enchant libraries on our tests for revscoring.

ores/score_caches/tests/test_redis.py
```diff
@@ -0,0 +1,46 @@
...
+                'ttl': 100}}}, 'redis')
+
+
+@pytest.mark.redis
+def test_lookup(cache):
+    cache.store(0.421, 'testwiki', 'damaging', 124)
+    assert cache.lookup('testwiki', 'damaging', 124) == 0.421
+
+
+class RedisMock():
```
adamwight: Move to its own module so it can be reused.

ores/task_tracker/tests/test_redis_task_tracker.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@pytest.mark.redis
+def test_release(task_tracker):
+    task_tracker.lock('fooo_release', 'value3')
+    assert task_tracker.release('fooo_release')
+    assert task_tracker.get_in_progress_task('fooo_release') is False
+
+
+class RedisMock():
```
adamwight: Please reuse from above.

ores/score_caches/tests/test_redis.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+
+from ..redis import Redis
```
adamwight: maybe import this as RedisScoreCache?  It gets confusing below, where we initialize Redis with a RedisMock.

ores/score_caches/tests/test_redis.py
adamwight: I agree that we should only mock here, but also fine with the integration tests as written.

ores/task_tracker/tests/test_redis_task_tracker.py
```diff
@@ -0,0 +1,67 @@
...
+    assert mock_task_tracker.lock('fooo', 'value1') is True
+
+
+def test_get_in_progress_task_mock(mock_task_tracker):
+    mock_task_tracker.lock('fooo_get', 'value2')
+    assert mock_task_tracker.get_in_progress_task('fooo_get') == 'value2'
+
+
+def test_release_mock(mock_task_tracker):
+    mock_task_tracker.lock('fooo_release', 'value3')
```
adamwight: Trivial thing—it's better to assign string constants like this to a variable because spelling is difficult.

ores/task_tracker/tests/test_task_tracker.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+
+from ..task_tracker import TaskTracker
+
+
+def test_lock():
+    task_tracker = TaskTracker()
+    with pytest.raises(NotImplementedError):
```
adamwight: Someone's shooting for the code coverage high score, I see :-p

ores/score_caches/tests/test_redis.py
Ladsgroup: In general I like to rename the class but for now, sure thing

ores/score_caches/tests/test_redis.py
Ladsgroup: Good idea!

ores/task_tracker/tests/test_redis_task_tracker.py
Ladsgroup: Done

ores/task_tracker/tests/test_redis_task_tracker.py
Ladsgroup: Good idea, I turned anything which was repeated more than twice to variable.

ores/task_tracker/tests/test_task_tracker.py
Ladsgroup: ¯\_(ツ)_/¯

PR #308, opened 2019-01-11
Make sure the ORES client always closes sockets

adamwight:
The requests `stream` parameter is probably unnecessary in our case, and can
cause issues where the socket isn't closed if an exception or other issue
causes us to not fully read the response.

Bug: T213582

### DIFF
--- a/ores/api.py
+++ b/ores/api.py
@@ -1,6 +1,8 @@
 """
 This module provides a :class:`ores.api.Session` class that can maintain a
-connection to an instance of ORES and efficiently generate scores.
+client connection to an instance of ORES and efficiently generate scores.
+
+Batching and parallelism are set by constructor arguments.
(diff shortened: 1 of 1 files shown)


ores/api.py
```diff
@@ -105,10 +107,11 @@ def _score_request(self, context, rev_ids, models):
...
                       .format(len(rev_ids)))
         start = time.time()
-        response = requests.get(url, params=params,
-                                headers=self.headers,
-                                verify=True, stream=True)
+
         try:
+            response = requests.get(url, params=params,
+                                    headers=self.headers,
+                                    verify=True)
```
halfak: ValueError should only apply to processing JSON.

adamwight: That's true, lemme back that out.

PR #309, opened 2019-01-11
Adds error handling and tests to score_revisions utility

halfak:


ores/api.py
```diff
@@ -81,22 +87,31 @@ def _score(self, context, models, rev_ids):
...
             # This loop loads all rev_id_batch's into the executor for
             # processing
             for rev_id_batch in chunked(rev_ids, self.batch_size):
                 rev_id_batch = list(rev_id_batch)
                 logging.debug("Starting batch of {0} revids"
                               .format(len(rev_id_batch)))
-                futures.append(executor.submit(self._score_request,
-                                               context, rev_id_batch,
-                                               models))
+                future_rev_ids.append((
```
adamwight: (The tuple is subtle.)

ores/api.py
```diff
@@ -81,22 +87,31 @@ def _score(self, context, models, rev_ids):
...
+                    executor.submit(self._score_request, context, rev_id_batch,
+                                    models),
+                    rev_id_batch))
```
adamwight: I think we can avoid this by pushing down the responsibility for stuffing error result cells.

ores/api.py
```diff
@@ -107,10 +122,9 @@ def _score_request(self, context, rev_ids, models):
         logging.debug("Sending score request for {0} revisions"
                       .format(len(rev_ids)))
         start = time.time()
-        response = requests.get(url, params=params,
-                                headers=self.headers,
-                                verify=True)
-
+        response = self._session.get(url, params=params,
```
adamwight: This is where an exception should be turned into JSON, probably with a `try` block to catch API errors.

ores/api.py
```diff
@@ -81,22 +87,31 @@ def _score(self, context, models, rev_ids):
...
+            for (future, rev_id_batch) in future_rev_ids:
+                try:
+                    for score in future.result():
+                        yield score
+                except RuntimeError as e:
+                    logger.warning(
+                        "An ORES scoring job failed with the following error:")
+                    logger.warning(traceback.format_exc())
+                    for rev_id in rev_id_batch:
+                        yield {m: {"error": e.args[0]}
```
adamwight: Responsibility more plausibly lies with `_score_request`, below.

ores/tests/test_api.py
```diff
@@ -1,6 +1,54 @@
 from .. import api
 
 
+class FakeSession:
+
+    def __init__(self, error_out=False):
+        self.error_out = error_out
+
+    def get(self, url, params, headers, verify=False, stream=False):
+        if self.error_out:
```
adamwight: This is just what MagicMock is for, you could define a mock to do this without extra state, in test_base_error below.

```
import mock
with mock.patch('requests.Session', return_value=FAKE_RESPONSE) as MockSession:
    api.Session("myhost")
```

ores/tests/test_api.py
```diff
@@ -1,6 +1,54 @@
...
+    assert 'error' in doc
+
+
 def test_session():
-    #  Ensures that the class can be constructed
-    api.Session("foo")
+    session1 = api.Session("myhost", session=FakeSession())
+    scores = list(session1.score("context", ["foo", "bar"], range(0, 100)))
+    assert len(scores) == 100
+    assert scores[0] == {"foo": {"score": "my score"},
```
adamwight: Great, thanks for including the happy path.  Nice tests!

ores/api.py
halfak: That's a good point.  I suppose we could have _score_request return an iterator of error docs in the case that the whole call fails.

ores/tests/test_api.py
halfak: Not sure I see what is going on here.  How does the MockSession know how what to return when .get() is called.  Maybe I need to study up more :|

halfak:
I don't see how the test parameter "session" can be avoided.

ores/api.py
halfak: Removed in most recent push.

ores/api.py
halfak: Fixed in most recent push.

ores/api.py
halfak: I think we should let errors that occur here bubble up.  I don't see much value in re-wrapping the error and we probably don't want to blindly handle a connection error or something like that.

ores/api.py
halfak: Good point.  {{done}}

ores/tests/test_api.py
halfak: Oh!  I guess we could monkey patch the `_session` variable in the test.  I feel conflicted about that though.

halfak:
...but I think I've addressed the rest of the concerns.

ores/tests/test_api.py
adamwight: It's just those mock lines I wrote above.  Either a mock or a fake object like you have is fine, but we should never include logic in production code which will only be used by tests.  A mock is also much simpler and easier to read because the special behavior is defined at the site of use.

To answer this question
> How does the MockSession know how what to return when .get() is called?

You would use the mock lines I pasted in order to provide an error from get(), or you can define the return_value as the happy-path function.  A mock also lets you assert that the method is called with expected arguments.

ores/api.py
adamwight: Just making sure I have this right—if we get a MediaWiki timeout error, it'll already be wrapped by lower-level functions and we won't encounter here, right?  It would be unfortunate to mark all revisions as errored.

I'm okay with the error map being built here though, at least it's in the right class.

halfak: Right.  This is a limitation of our batching strategy.  When we're batching 50 revisions together, any API failure will cause the scoring of all 50 revisions to fail.

adamwight:
This will work just fine, although I still recommend dropping the "session" parameter in the future as YAGNI.

Here's an article which may make mock objects joyful for you: https://martinfowler.com/articles/mocksArentStubs.html

PR #314, opened 2019-02-08
Another try to run travis tests with docker

Ladsgroup:
Bug: T183214

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,24 @@
 language: python
 sudo: true
-python:
-  - "3.5"
-addons:
-  apt:
-    packages:
-    - g++
-    - gfortran
-    - libblas-dev
-    - liblapack-dev
-    - libopenblas-dev
-    - python3-dev
-    - enchant
-    - redis-server
-
+env:
+  - DOCKER_COMPOSE_VERSION=1.17.1
+services:
+  - docker
 before_install:
-  - npm install -g eslint
-  - sudo sed -e 's/^bind.*/bind 127.0.0.1/' /etc/redis/redis.conf > redis.conf
-  - sudo mv redis.conf /etc/redis
-  - sudo service redis-server start
-  - echo PING | nc localhost 6379
+  - sudo rm /usr/local/bin/docker-compose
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,43 +1,24 @@
...
+env:
+  - DOCKER_COMPOSE_VERSION=1.17.1
+services:
+  - docker
 before_install:
-  - npm install -g eslint
-  - sudo sed -e 's/^bind.*/bind 127.0.0.1/' /etc/redis/redis.conf > redis.conf
-  - sudo mv redis.conf /etc/redis
-  - sudo service redis-server start
-  - echo PING | nc localhost 6379
```
halfak: What's going on here?  Looks like this needs a comment.

.travis.yml
```diff
@@ -1,43 +1,24 @@
...
-  - flake8 .
-  - eslint .
-  - ./utility applications.wsgi --config ./config/celery-test/ &
-  - ./utility applications.celery --config ./config/celery-test/ &
-  - sleep 5
-  - ./utility test_api http://localhost:8080 --debug
-  - py.test --cov=ores -vv
+  - docker-compose up -d
+  - docker exec ores_ores-worker_1 flake8 /ores
+#  - eslint .
```
halfak: How come we're not running this?

.travis.yml
Ladsgroup: All of this is being handled in the docker-compose. Docker compose brings up the redis and all of the needed dependencies.

.travis.yml
Ladsgroup: That's the next thing in my list to tackle. It's not hard though. I just need to do it.

Ladsgroup: Also I want to mention this is linting for js/css. We don't have much of it in ores. I will fix it though.

.travis.yml
halfak: I think it's showing the wrong side of the diff.  See, the code you added removes docker-compose and re-downloads it.  Why?

Ladsgroup: aha, docker compose is not installed by default, it needs to be installed: https://docs.travis-ci.com/user/docker/#using-docker-compose

Ladsgroup: I removed that part completely, it's not needed.

halfak:
OK.  It looks like we need a follow-up task for eslint, but otherwise, this looks good.

PR #315, opened 2019-02-11
Upgrade to flask 1.0

Ladsgroup:
Bug: T212002

### DIFF
--- a/ores/applications/wsgi.py
+++ b/ores/applications/wsgi.py
@@ -46,7 +46,7 @@ def main(argv=None):
 def run(host, port, processes, **kwargs):
     application = build(**kwargs)
     application.debug = True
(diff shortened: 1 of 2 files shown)


ores/applications/wsgi.py
```diff
@@ -46,7 +46,7 @@ def main(argv=None):
 def run(host, port, processes, **kwargs):
     application = build(**kwargs)
     application.debug = True
-    application.run(host=host, port=port, processes=processes, debug=True)
+    application.run(host=host, port=port, processes=processes, threaded=False, debug=True)
```
halfak: What a strange thing.  We need to explicitly say not to be threaded even when we're specifying > 1 processes?  Oh well.  Good catch.

PR #318, opened 2019-02-24
More cleanup of docker parts

Ladsgroup:
Bug: T215906

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,6 @@ RUN pip install wheel
 RUN pip install nltk
 RUN pip install -r /ores/requirements.txt
 RUN pip install -r /ores/test-requirements.txt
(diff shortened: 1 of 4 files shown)


Ladsgroup:
The failure is because of a mistake in revscoring that is fixed by https://github.com/wikimedia/revscoring/pull/434

Dockerfile
```diff
@@ -18,7 +18,6 @@ RUN pip install wheel
 RUN pip install nltk
 RUN pip install -r /ores/requirements.txt
 RUN pip install -r /ores/test-requirements.txt
-RUN pip install codecov pytest-cov
```
halfak: We don't need to install this anymore?

Ladsgroup: It's in test-requirements.txt and installed above.

PR #319, opened 2019-02-25
Limit root_caches to only contain target root dependencies. 

Ladsgroup:
Bug: T214841

### DIFF
--- a/ores/scoring_context.py
+++ b/ores/scoring_context.py
@@ -170,10 +170,12 @@ class ScoringContext(dict):
 
         # Check each extraction for errors
         errors = {}
-        for rev_id, (error, _) in zip(rev_ids, error_root_vals):
+        for rev_id, (error, values) in zip(rev_ids, error_root_vals):
             if error is not None:
                 errors[rev_id] = error
                 del root_caches[rev_id]
+            else:
+                root_caches[rev_id] = dict(zip(root_datasources, values))
         logger.debug("Extracted root datasources for {0}:{1}:{2} in {3} secs"
                      .format(self.name, set(model_names), rev_ids,
                              round(time.time() - start, 3)))
--- a/ores/scoring_context.py
+++ b/ores/scoring_context.py
@@ -3,7 +3,6 @@ import time
 from hashlib import sha1
 
 from revscoring import Model, dependencies
-from revscoring.datasources import Datasource
 from revscoring.extractors import Extractor
 from revscoring.features import trim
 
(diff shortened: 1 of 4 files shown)


halfak:
This doesn't quite do what it needs to.  The root_caches need to include the injected_cache when extract is called.  So if someone specifies a root datasource, it is then used during extraction.

ores/scoring_systems/tests/test_single_thread.py
```diff
@@ -11,11 +11,11 @@ def test_score():
 
 
 def test_single_thread():
-    scoring_system = SingleThread({'fakewiki': fakewiki}, timeout=1)
+    scoring_system = SingleThread({'fakewiki': fakewiki}, timeout=1.0)
```
halfak: Should have been a floating point number

ores/scoring_systems/tests/test_single_thread.py
```diff
@@ -11,11 +11,11 @@ def test_score():
...
 
     response = scoring_system.score(
         ScoreRequest("fakewiki", [1], ["fake"],
-                     injection_caches={1: {wait_time: 2}}))
+                     injection_caches={1: {wait_time: 2.0}}))
```
halfak: Needs to be floating point

ores/scoring_systems/tests/util.py
```diff
@@ -34,8 +35,9 @@ def score(self, feature_values):
 
 class FakeExtractor(Extractor):
 
-    def extract(self, rev_ids, features, *args, **kwargs):
-        return [(None, [0 for feature in features]) for rev_id in rev_ids]
+    def extract(self, rev_ids, features, *args, caches={}, **kwargs):
+        return [(None, list(solve(features, cache=caches.get(rev_id, {}))))
+                for rev_id in rev_ids]
```
halfak: The old version here was doing literally nothing.  We need to it abide by the extraction pattern for the test to pass.

ores/scoring_context.py
```diff
@@ -152,28 +151,32 @@ def extract_root_dependency_caches(
...
         for rev_id in rev_ids:
             injection_cache = injection_caches.get(rev_id, {}) \
                               if injection_caches is not None else {}
-            root_caches[rev_id] = dict(injection_cache.items())
+            _injection_caches[rev_id] = dict(injection_cache.items())
 
         # Find our root datasources
         root_datasources = \
             list(set(self._generate_root_datasources(model_names)))
+        print(root_datasources, _injection_caches)
```
Ladsgroup: Leftover from debug

halfak: Woops!  Good catch!

Ladsgroup:
$ docker-compose exec ores-worker flake8 /ores
/ores/ores/scoring_context.py:6:1: F401 'revscoring.datasources.Datasource' imported but unused

halfak:
Build failing seems to be unrelated.  Looks like our tests may be broken due to a version conflict.

halfak:
Weird.  A strange new failure showed up.  Digging into it now.

PR #320, opened 2019-03-27
Move tests to outside of the production code

Ladsgroup:
It's standard in python

Bug: T219360


halfak:
I rebased on master with redis version fix.

PR #322, opened 2019-03-27
Fix order of imports

Ladsgroup:
Using isort library


halfak:
I re-ran isort to make sure I didn't break anything after rebasing and addressing conflicts.  But it made a bunch of additional changes, so I just committed those too.  Here's the command I ran: 

```
isort --recursive ./tests/ -p ores
```

PR #327, opened 2019-04-29
Fixes root features bug

halfak:
T222121


halfak:
I don't like it, but given the status of this issue, I'm self-merging.

PR #328, opened 2019-07-23
Switch to meta.schema from meta.topic in precache

halfak:
https://phabricator.wikimedia.org/T228689

### DIFF
--- a/ores/wsgi/util.py
+++ b/ores/wsgi/util.py
@@ -1,6 +1,5 @@
 import json
 import logging
-import re
 
 from flask_jsonpify import jsonify as flask_jsonify
 
@@ -182,8 +181,7 @@ def build_event_set(event):
     uses internally.
     """
     event_set = set()
-    if re.match(r"([^\.]+.)?mediawiki\.revision-create$",
-                event['meta']['topic']):
+    if event['meta']['stream'] == "mediawiki.revision-create":
         event_set.add('edit')
 
         user_groups = event.get('performer', {}).get('user_groups', [])
--- a/tests/wsgi/tests/test_util.py
+++ b/tests/wsgi/tests/test_util.py
@@ -1,6 +1,5 @@
 import flask
 import pytest
-
 from ores.applications.wsgi import build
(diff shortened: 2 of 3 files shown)


tests/wsgi/tests/test_util.py
```diff
@@ -72,7 +71,10 @@ def test_build_precache_map():
 def test_event():
     event = {
         "database": "enwiki",
-        "meta": {"topic": "mediawiki.revision-create"},
+        "meta": {
+            "topic": "eqiad.mediawiki.revision-create",
+            "schema": "mediawiki.revision-create"
```
ottomata: The schema field is now at `$schema` top level.  The event will have these like:
```
$schema: /mediawiki/revision/create/1.0.0
meta: {
  stream: mediawiki.revision-create
  ...
}

File: tests/wsgi/tests/test_util.py [hunk 2]
```diff
@@ -72,7 +71,10 @@ def test_build_precache_map():
 def test_event():
     event = {
         "database": "enwiki",
-        "meta": {"topic": "mediawiki.revision-create"},
+        "meta": {
+            "topic": "eqiad.mediawiki.revision-create",
```
Pchelolo: There will be no 'topic' in the event at all, you can delete this from the test.

ores/wsgi/util.py
```diff
@@ -182,8 +182,7 @@ def build_event_set(event):
     uses internally.
     """
     event_set = set()
-    if re.match(r"([^\.]+.)?mediawiki\.revision-create$",
-                event['meta']['topic']):
+    if event['meta']['schema'] == "mediawiki.revision-create":
```
Pchelolo: I'm wondering if that check is needed at all. Changeprop is configured to only update ORES on `revision-create`, so nothing else could be sent here.

halfak: Eventually, we'll want to be able to precache based on other types of events.  So I'd like to keep it here.  Either way, it's good practice to check our assumption explicitly.

Pchelolo:
LGTM

PR #333, opened 2019-10-24
Limit rev_ids in an ORES request to 50

kevinbazira:
https://phabricator.wikimedia.org/T232005

### DIFF
--- a/ores/wsgi/routes/v3/util.py
+++ b/ores/wsgi/routes/v3/util.py
@@ -98,8 +98,12 @@ def build_v3_context_model_map(score_request, scoring_system):
 def process_score_request(score_request, scoring_system):
     score_request.model_info = score_request.model_info or ['version']
     try:
-        score_response = scoring_system.score(score_request)
-        return format_v3_score_response(score_response)
+        if len(score_request.rev_ids) > 50:
+            return responses.bad_request(
+                "Too many values for 'revids' parameter.  Max of 50.")
+        else:
+            score_response = scoring_system.score(score_request)
+            return format_v3_score_response(score_response)
     except errors.ScoreProcessorOverloaded:
         scoring_system.metrics_collector.response_made(
                 responses.SERVER_OVERLOADED, score_request)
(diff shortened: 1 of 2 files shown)


ores/wsgi/routes/v3/util.py
```diff
@@ -98,8 +98,13 @@ def build_v3_context_model_map(score_request, scoring_system):
 def process_score_request(score_request, scoring_system):
     score_request.model_info = score_request.model_info or ['version']
     try:
-        score_response = scoring_system.score(score_request)
-        return format_v3_score_response(score_response)
+        if len(score_request.rev_ids) > 50:
+            raise ValueError('Should have 50 rev_ids or less in a request.')
```
halfak: It looks like we want to return an error response.  

See the following line below as an example:
```
return responses.not_found("No scorers available for {0}"
                           .format(e))
```

It looks like `bad_request` is probably what you want to return.  See https://github.com/wikimedia/ores/blob/435e5cc8a4078e35bfbbbd4f2b20a530c9e87db7/ores/wsgi/responses.py#L30 for the definition

ores/wsgi/routes/v3/util.py
```diff
@@ -98,8 +98,13 @@ def build_v3_context_model_map(score_request, scoring_system):
...
+        else:
+            score_response = scoring_system.score(score_request)
+            return format_v3_score_response(score_response)
+    except ValueError as e:
+        return e
```
halfak: I'm not sure what this ValueError is catching

ores/wsgi/routes/v3/util.py
```diff
@@ -98,8 +98,12 @@ def build_v3_context_model_map(score_request, scoring_system):
 def process_score_request(score_request, scoring_system):
     score_request.model_info = score_request.model_info or ['version']
     try:
-        score_response = scoring_system.score(score_request)
-        return format_v3_score_response(score_response)
+        if len(score_request.rev_ids) > 50:
+            return responses.bad_request(
+                "Should have 50 rev_ids or less in a request.")
```
halfak: Let's change this error message to: 
```
"Too many values for 'revids' parameter.  Max of 50."
```

This clearly references the 'revids' parameter -- which is what the user will see.

PR #335, opened 2019-12-19
Adds a test to check on our build_event_set method.

halfak:
I wrote this test while I was checking on an issue.  We might as well include it in the suite.

### DIFF
--- a/tests/wsgi/tests/test_util.py
+++ b/tests/wsgi/tests/test_util.py
@@ -1,9 +1,9 @@
 import flask
 import pytest
-
 from ores.applications.wsgi import build
 from ores.scoring_systems.scoring_system import ScoringSystem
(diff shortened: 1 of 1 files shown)


tests/wsgi/tests/test_util.py
```diff
@@ -88,6 +88,42 @@ def test_build_precache_map():
     precache_map = build_precache_map(config)
     assert precache_map == {'enwiki': {'edit': {'revid', 'goodfaith'}}}
 
+def test_build_event_set():
```
accraze: Looks like the flake8 check is failing on TravisCI here:
```
/ores/tests/wsgi/tests/test_util.py:91:1: E302 expected 2 blank lines, found 1
```

Add another line before this test function and you should be good to go.

PR #342, opened 2020-04-24
Switches default config to use no cache so pylru is not required.

halfak:
Addresses https://github.com/wikimedia/ores/issues/166

### DIFF
--- a/ores/score_caches/empty.py
+++ b/ores/score_caches/empty.py
@@ -8,3 +8,7 @@ class Empty(ScoreCache):
 
     def store(self, *args, **kwargs):
         return None
+
+    @classmethod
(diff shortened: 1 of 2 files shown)


ores/score_caches/empty.py
```diff
@@ -8,3 +8,9 @@ def lookup(self, *args, **kwargs):
 
     def store(self, *args, **kwargs):
         return None
+
+    @classmethod
+    def from_config(cls, config, name, section_key="score_caches"):
+        section = config[section_key][name]
```
accraze, line 14: Flake8 error: ores/score_caches/empty.py:14:9: F841 local variable 'section' is assigned to but never used

halfak: Thanks!  Got it.

PR #345, opened 2020-05-18
[WIP] Blubberfile & Python 3.7

accraze:
This PR contains a [Blubberfile](https://wikitech.wikimedia.org/wiki/Blubber) for use in the Wikimedia Deployment pipeline. This image relies on Py3.7, so there were a few changes made to support the upgrade (bump celery version and some test deps).

A couple of open questions:
* How should we handle the Redis dependency with the test variant? (should we mock or spin up a small instance?)
* The scipy stack is heavy and takes up a good amount of space (500mb+) should we build our own base image that has the pre-compiled binaries?


accraze:
Closing due to splitting the work into two separate PRs:

- **Python 3.7 changes:** #348 
- **Blubberfile & PipelineLib config:** #349

PR #348, opened 2020-07-02
Build: Python 3.7 base image & layer caching

accraze:
This PR updates the Dockerfile to use a Python3.7 base image, introduces layer caching to speed up build times and also has a change to reduce the overall image size.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.5-slim-stretch
+FROM python:3.7-slim-stretch
 
 RUN apt-get update && apt-get install -y \
     g++ \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,23 +1,21 @@
...
+COPY test-requirements.txt /ores/test-requirements.txt
+RUN pip install pip --upgrade && pip install wheel && pip install nltk \
+    && pip install -r /ores/requirements.txt \
+    && pip install -r /ores/test-requirements.txt \
+    && python -m nltk.downloader stopwords
 
 COPY . /ores
 WORKDIR /ores
 
-RUN pip install pip --upgrade
```
accraze, line 16: Each of these `RUN` commands adds an additional layer to the image and increases the overall size. We can avoid this by concatenating these commands in a single layer instead.

Dockerfile
```diff
@@ -1,23 +1,21 @@
...
 RUN apt-get update && apt-get install -y \
     g++ \
     gfortran \
     libblas-dev \
     liblapack-dev \
     libopenblas-dev \
-    python3-dev \
     enchant \
     build-essential
+COPY requirements.txt /ores/requirements.txt
```
accraze, line 11: Requirements tend to change less often than the source code. If we `COPY` the requirements before the source code, then we can speed up build time by avoiding re-installing all the dependencies.

PR #354, opened 2021-11-27
Fix: replaces in wsgi score.js should be global

perryprog:
Pretty simple fix so I didn't go with a phab ticket. Currently the [WSGI UI](https://ores.wikimedia.org/ui/) (which I suspect very few people use, so this is very low impact) states it allows for entering multiple rev IDs that are comma separated, but if you attempt to do so with three or more, it errors due to "1,2,3" being changed to "1|2,3" instead of what the server expects which is "1|2|3". I tested it locally and it worked as expected.

(Force-pushed to correct a typo in a commit message.)

### DIFF
--- a/ores/wsgi/static/js/scorer.js
+++ b/ores/wsgi/static/js/scorer.js
@@ -80,7 +80,7 @@ function createTable( data ) {
 }
 
 function getResults() {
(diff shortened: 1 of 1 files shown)


he7d3r (approved):

accraze (approved):
:eyes: :+1:

